PR #1, opened 2018-02-22
Fix bootstrapping problem

isuruf:


.travis.yml
```diff
@@ -15,7 +15,7 @@ env:
 before_install:
     # Fast finish the PR.
     - |
-      (cat recipe/ff_ci_pr_build.py | \
+      (curl https://raw.githubusercontent.com/conda-forge/conda-forge-build-setup-feedstock/master/recipe/ff_ci_pr_build.py | \
```
jakirkham: This should still be ok. It's just using the file already in `recipe` independently of installing the package.

.travis.yml
```diff
@@ -48,10 +48,9 @@ install:
       conda config --add channels conda-forge
       conda config --set show_channel_urls true
       conda install --yes --quiet conda-forge-build-setup
-      # Overriding global conda-forge-build-setup with local copy.
-      source recipe/run_conda_forge_build_setup_osx
+      source run_conda_forge_build_setup
```
jakirkham: I think we manually need to install `conda-build`.

jakirkham:
So I think all we need to do is install `conda-build` in the CI scripts manually. We could even keep a step like this in the scripts so there won't be a bootstrapping problem in the future either.

.travis.yml
```diff
@@ -48,10 +48,12 @@ install:
...
+      conda update -n root --yes --quiet conda conda-env conda-build
+      conda install -n root --yes --quiet jinja2 anaconda-client
+      conda install -n root --yes --quiet conda-build=2
 
 script:
   - conda build ./recipe
 
-  - recipe/upload_or_check_non_existence.py ./recipe conda-forge --channel=main
+  - upload_or_check_non_existence ./recipe conda-forge --channel=main
```
jakirkham: Same story here. This should be fine to leave as is.

isuruf: Nope. Need conda-build=3 for this recipe

jakirkham: Right, because the upload script only works with cb3. Sorry. Please ignore my comment then.

jakirkham:
Good to merge?

isuruf:
Thanks for the review

PR #4, opened 2018-02-28
MNT: Re-rendered with conda-smithy 2.4.5.post.dev147 and pinning 2018…

isuruf:
….02.20


isuruf:
Looks like `CONDA_BUILD_SYSROOT` is not passed through by conda-build. We can generate an additional conda-build config file here and pass it to conda-build.

cc @msarahan

jakirkham:
Thanks for testing @isuruf.

Wonder if we include environment variables into `conda_build_config.yaml` somehow. Maybe by using jinja?

recipe/run_conda_forge_build_setup_osx
```diff
@@ -14,14 +14,16 @@ export MACOSX_DEPLOYMENT_TARGET="10.9"
...
-if [ -f "${CONDA_BUILD_SYSROOT}" ]
+if [ -d "${CONDA_BUILD_SYSROOT}" ]
 then
     echo "Found CONDA_BUILD_SYSROOT: ${CONDA_BUILD_SYSROOT}"
 else
     echo "Missing CONDA_BUILD_SYSROOT: ${CONDA_BUILD_SYSROOT}"
     exit 1
 fi
 
+ln -s ${CONDA_BUILD_SYSROOT} /opt/MacOSX10.9.sdk
```
isuruf: @jakirkham, what do you think about this?

jakirkham: Probably works. Don't know if it would matter to use a symlink or a hardlink. Other thing to check would be how the rpaths or set (in case this might confuse matters there).

OTOH maybe it isn't worth the trouble and we should just hard code the whole value in `conda-forge-pinning`. FWIW this is what it would be ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/16 ).

Would also note that the new compilers pick up this setting by default anyways. Maybe nothing else needs them. Still it might make sense for `conda-build` to provide this default as well ( https://github.com/conda/conda-build/issues/2719 ).

recipe/run_conda_forge_build_setup_osx
```diff
@@ -14,14 +14,20 @@ export MACOSX_DEPLOYMENT_TARGET="10.9"
 
 export CONDA_BUILD_SYSROOT="$(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk"
 
-if [ -f "${CONDA_BUILD_SYSROOT}" ]
+if [ -d "${CONDA_BUILD_SYSROOT}" ]
```
jakirkham: Sorry about this. Thanks for fixing.

recipe/run_conda_forge_build_setup_osx
```diff
@@ -14,14 +14,20 @@ export MACOSX_DEPLOYMENT_TARGET="10.9"
...
     echo "Found CONDA_BUILD_SYSROOT: ${CONDA_BUILD_SYSROOT}"
 else
     echo "Missing CONDA_BUILD_SYSROOT: ${CONDA_BUILD_SYSROOT}"
     exit 1
 fi
 
+if [ ! -z "$CONFIG" ]; then
+    echo "CONDA_BUILD_SYSROOT:" >> ./.ci_support/${CONFIG}.yaml
+    echo "- ${CONDA_BUILD_SYSROOT}" >> ./.ci_support/${CONFIG}.yaml
+    cat ./.ci_support/${CONFIG}.yaml
```
isuruf: @jakirkham, what do you think about this?

jakirkham: Interesting! This is a nice idea. :)

So in this case we would drop `CONDA_BUILD_SYSROOT` from the `conda-forge-pinning`?

jakirkham: Sorry. I think we posted around the same time. Comment below.

In general, think this is a nice idea.

isuruf: > So in this case we would drop CONDA_BUILD_SYSROOT from the conda-forge-pinning?

Yes

recipe/run_conda_forge_build_setup_osx
```diff
@@ -14,14 +14,20 @@ export MACOSX_DEPLOYMENT_TARGET="10.9"
...
+if [ -d "${CONDA_BUILD_SYSROOT}" ]
 then
     echo "Found CONDA_BUILD_SYSROOT: ${CONDA_BUILD_SYSROOT}"
 else
     echo "Missing CONDA_BUILD_SYSROOT: ${CONDA_BUILD_SYSROOT}"
     exit 1
 fi
 
+if [ ! -z "$CONFIG" ]; then
+    echo "CONDA_BUILD_SYSROOT:" >> ./.ci_support/${CONFIG}.yaml
```
jakirkham: We might want a blank line or two before this for good measure.

.appveyor.yml
```diff
@@ -27,14 +23,17 @@ install:
...
 
     # Add path, activate `conda` and update conda.
-    - cmd: call %CONDA_INSTALL_LOCN%\Scripts\activate.bat
-    - cmd: conda.exe update --yes --quiet conda
+    # Around November of 2017, Appveyor changed Miniconda3-x64 from an old 3.4 install to a
+    #     symlink to the current version.
+    - cmd: call C:\Miniconda3-x64\Scripts\activate.bat
+
+    # workaround for https://github.com/conda/conda/issues/6057
+    - cmd: conda.exe update conda
```
jakirkham: Needs `--yes --quiet`. Currently it is timing out without these.

jakirkham: Addressing in PR ( https://github.com/conda-forge/conda-smithy/pull/696 ).

jakirkham:
Pushed a fresh re-rendering (as it should fix Windows). Hope that is ok.

jakirkham:
Passes. Ok to merge?

PR #7, opened 2018-03-05
Support circleci OSX

isuruf:
* [x] Bump the build number (if the version is unchanged)
* [x] [Re-render]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`
* [x] Ensure the license file is being packaged.

### DIFF
--- a/recipe/run_conda_forge_build_setup_osx
+++ b/recipe/run_conda_forge_build_setup_osx
@@ -12,6 +12,13 @@ export PYTHONUNBUFFERED=1
 #   way is working well enough.
 export MACOSX_DEPLOYMENT_TARGET="10.9"
 
+if [ "$CIRCLECI" == "true" ]; then
+    echo "Downloading sdk"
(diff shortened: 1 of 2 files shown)


recipe/run_conda_forge_build_setup_osx
```diff
@@ -12,6 +12,13 @@ export PYTHONUNBUFFERED=1
 #   way is working well enough.
 export MACOSX_DEPLOYMENT_TARGET="10.9"
 
+if [ "$CIRCLECI" == "true" ]; then
+    echo "Downloading sdk"
+    curl -L -O https://github.com/phracker/MacOSX-SDKs/releases/download/10.13/MacOSX10.9.sdk.tar.xz
```
jakirkham: Another option would be to package our own copy of the SDK. This way we could just always install that conda package and use it regardless of what CI we are on.

xref: https://github.com/conda-forge/staged-recipes/pull/5382

isuruf: Since we don't know whether we can distribute it as a conda package, I'd go with this option.

Issue #14, opened 2018-05-01
Sporadic cleanup issues on AppVeyor after upload


jakirkham:
As [noted in this comment]( https://github.com/conda-forge/r-base-feedstock/pull/39#issuecomment-373042970 ) and [again in this comment]( https://github.com/conda-forge/conda-feedstock/pull/46#issuecomment-385570804 ), it seems that sometimes we lack permissions to do a little bit of cleanup on AppVeyor (possibly due to resource contention), which causes the build to fail after uploading packages. Since this appears to be occurring after upload it isn't actually messing the uploads, but it does give one pause. Restarting seems to make it go away, but it does show up too often for comfort and it does make users nervous. Maybe sleeping for a bit before cleaning up would avoid this issue. There may be better fixes though.

```python
Traceback (most recent call last):
  File "C:\Miniconda36-x64\Scripts\upload_or_check_non_existence.py", line 131, in <module>
    main()
  File "C:\Miniconda36-x64\Scripts\upload_or_check_non_existence.py", line 121, in main
    upload(cli, fname, owner, channel)
  File "C:\Miniconda36-x64\Scripts\upload_or_check_non_existence.py", line 68, in upload
    env=os.environ)
  File "C:\Miniconda36-x64\lib\contextlib.py", line 88, in __exit__
    next(self.gen)
  File "C:\Miniconda36-x64\Scripts\upload_or_check_non_existence.py", line 27, in get_temp_token
    shutil.rmtree(dn)
  File "C:\Miniconda36-x64\lib\shutil.py", line 494, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\Miniconda36-x64\lib\shutil.py", line 389, in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
  File "C:\Miniconda36-x64\lib\shutil.py", line 387, in _rmtree_unsafe
    os.unlink(fullname)
PermissionError: [WinError 5] Access is denied: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\tmphmihw0df\\binstar.token'
```

ref: https://ci.appveyor.com/project/conda-forge/conda-feedstock/build/1.0.122
ref: https://ci.appveyor.com/project/conda-forge/r-base-feedstock/build/1.0.90

nehaljwani:
But not all pkgs have been uploaded! Current status for windows:

[...]

PR #17, opened 2018-05-21
AppVeyor: Workaround for PermissionError during shutil.rmtree

mbargull:
Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`

*Might* resolve gh-14 or at least make it occur less frequently.

It looks like it's very possible that the cleanup `shutil.rmtree(dn)` call fails on Windows due to open file handles or such. Sleeping a short while and trying again might help. Seems like `conda-build` suffers the same fate, see https://github.com/conda-forge/conda-feedstock/pull/52.
**NOTE**: That `0.1` second duration is chosen arbitrarily; maybe it could be shorter, maybe it should be longer, I don't know.

### DIFF
--- a/recipe/upload_or_check_non_existence.py
+++ b/recipe/upload_or_check_non_existence.py
@@ -9,12 +9,15 @@ import shutil
 import subprocess
 import sys
 import tempfile
+import time
 
 from binstar_client.utils import get_server_api
 import binstar_client.errors
 from conda_build.conda_interface import subdir as conda_subdir
 from conda_build.conda_interface import get_index
 import conda_build.api
+
+on_win = (sys.platform == "win32")
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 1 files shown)


mbargull:
If you think this change makes sense, should this get a build or a version number bump? I assume the latter?

scopatz:
I believe that @CJ-Wright has seen this in a few cases and has a more robust solution for windows, somewhere.  Where was that lying?

mbargull:
> and has a more robust solution for windows, somewhere

That'd be great! I know @kalefranz has been (and I think still is?) battling those beasts for `conda`, too. (I deliberately ignored those issues and his progress on this, though -- sorry Kale :wink:.)
Be sure to let us know if you got something on this, @CJ-Wright, please :smile:.

CJ-Wright:
I think you are looking for https://github.com/regro/rever/blob/master/tests/conftest.py#L44

scopatz:
Yup that is it!

PR #30, opened 2018-10-02
fix bash

ocefpaf:
@mariusvniekerk can you take a look and merge?

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "2.0.7" %}
+{% set version = "2.0.8" %}
 
 package:
   name: {{ name|lower }}


recipe/run_conda_forge_build_setup_osx
```diff
@@ -12,7 +12,7 @@ export PYTHONUNBUFFERED=1
 #   way is working well enough.
 export MACOSX_DEPLOYMENT_TARGET="10.9"
 
-if [ "$CIRCLECI" == "true" || "$OSX_FORCE_SDK_DOWNLOAD" == "1" ]; then
+if [[ "$CIRCLECI" == "true" || "$OSX_FORCE_SDK_DOWNLOAD" == "1" ]]; then
```
mbargull: Should be fine. POSIX-y alternative would be
```sh
if [ "$CIRCLECI" == "true" ] || [ "$OSX_FORCE_SDK_DOWNLOAD" == "1" ]; then
```
IDK if there is any preference for the one or the other here, but I guess it shouldn't matter since we have `#!/bin/bash` anyway.

Can you also bump the version number?

ocefpaf: > Can you also bump the version number?

Done!

mbargull: Yes, the current `[ ... || ... ]` is broken.
`[[ ... || ... ]]` or `[ ... ] || [ ... ]` are the working alternatives is what I meant to say :wink:.

jakirkham: Doesn't look like CircleCI was using @mbargull's suggestion. That said, happy to go with the current fix (we force `bash` in this script anyways 😉).

ocefpaf: Hard to read the diff when your eyes are trained to the the first and last bracket ;-p

mbargull:
(erm, "it's blocking macOS builds"... not really since `2.0.7` is already moved to the `broken` label -- well, doesn't harm anyway :wink: )

PR #31, opened 2018-10-18
Download mac sdk if not present

minrk:
Rather than checking `if circleci`, check if the sdk is needed. This prepares for updating the Travis xcode image to a more recent version that will not have the aged 10.9 SDK available.

Additionally:

- inherit rather than override MACOSX_DEPLOYMENT_TARGET, to avoid overriding the new way to set this with the old way
- use plutil instead of sed to rewrite MinimumSDKVersion (more future-proof, works for binary plists, which current versions of xcode use, isn't sensitive to previous value)

The Travis image needs to be updated pretty soon: https://github.com/conda-forge/conda-forge.github.io/issues/641

### DIFF
--- a/recipe/run_conda_forge_build_setup_osx
+++ b/recipe/run_conda_forge_build_setup_osx
@@ -6,11 +6,9 @@ export CPU_COUNT=2
 
 export PYTHONUNBUFFERED=1
 
-# TODO: this has moved to conda_build_config.yaml, but we are keep it here for a little while, until
-#   the dust settles on using conda-build 3.
-#   This now only sets the value when it is undefined,
-#   which will eventually become entirely redundant.
-#   Remove this value when the new way is working well enough.
+# deployment target should be set by conda_build_config.yaml (default in conda-forge-pinning).
+# The default here will only be used when that is undefined,
+# which should only be recipes still using conda-build 2.
 export MACOSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET:-10.9}
 
 export CONDA_BUILD_SYSROOT="$(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk"
--- a/recipe/run_conda_forge_build_setup_osx
+++ b/recipe/run_conda_forge_build_setup_osx
@@ -7,19 +7,21 @@ export CPU_COUNT=2
 export PYTHONUNBUFFERED=1
 
 # TODO: this has moved to conda_build_config.yaml, but we are keep it here for a little while, until
-#   the dust settles on using conda-build 3.  If this value differs from the one in conda_build_config.yaml,
-#   things might get confusing, so make sure to keep them the same, and remove this value when the new
-#   way is working well enough.
-export MACOSX_DEPLOYMENT_TARGET="10.9"
-
-if [[ "$CIRCLECI" == "true" || "$OSX_FORCE_SDK_DOWNLOAD" == "1" ]]; then
-    echo "Downloading sdk"
-    curl -L -O https://github.com/phracker/MacOSX-SDKs/releases/download/10.13/MacOSX10.9.sdk.tar.xz
-    tar -xf MacOSX10.9.sdk.tar.xz -C $(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/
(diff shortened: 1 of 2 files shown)


recipe/run_conda_forge_build_setup_osx
```diff
@@ -7,20 +7,22 @@ export CPU_COUNT=2
...
 
 export CONDA_BUILD_SYSROOT="$(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk"
 
+if [[ ! -d ${CONDA_BUILD_SYSROOT} || -z "$OSX_FORCE_SDK_DOWNLOAD" == "1" ]]; then
+    echo "Downloading ${MACOSX_DEPLOYMENT_TARGET} sdk"
+    curl -L -O https://github.com/phracker/MacOSX-SDKs/releases/download/10.13/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk.tar.xz
+    tar -xf MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk.tar.xz -C "$(dirname "$CONDA_BUILD_SYSROOT")"
+    # set minimum sdk version to our target
+    plutil -replace MinimumSDKVersion -string ${MACOSX_DEPLOYMENT_TARGET} $(xcode-select -p)/Platforms/MacOSX.platform/Info.plist
+fi
```
scopatz: @mariusvniekerk - is this something that we do on azure?

scopatz:
Thanks for putting this in @minrk! I think that this is a good change to have.  seems like there are some syntax errors on Travis:

```
Executing transaction: ...working... done
recipe/run_conda_forge_build_setup_osx: line 18: syntax error in conditional expression
recipe/run_conda_forge_build_setup_osx: line 18: syntax error near `=='
recipe/run_conda_forge_build_setup_osx: line 18: `if [[ ! -d ${CONDA_BUILD_SYSROOT} || -z "$OSX_FORCE_SDK_DOWNLOAD" == "1" ]]; then'
```

minrk:
@scopatz thanks! Fixed.

recipe/run_conda_forge_build_setup_osx
```diff
@@ -7,20 +7,22 @@ export CPU_COUNT=2
...
-    echo "Downloading sdk"
-    curl -L -O https://github.com/phracker/MacOSX-SDKs/releases/download/10.13/MacOSX10.9.sdk.tar.xz
-    tar -xf MacOSX10.9.sdk.tar.xz -C $(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/
-    sed -i '' "s|<string>10.11</string>|<string>10.9</string>|g" $(xcode-select -p)/Platforms/MacOSX.platform/Info.plist
-fi
+#   the dust settles on using conda-build 3.
+#   This now only sets the value when it is undefined,
+#   which will eventually become entirely redundant.
+#   Remove this value when the new way is working well enough.
+export MACOSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET:-10.9}
```
jakirkham: Sorry I think I'm missing something, but what is the new way? Would it be possible to link to relevant docs, issue(s), or PR(s) in the comment to provide additional information? The fewer links and the closer to proper docs the better, but I understand if that isn't totally possible.

jakirkham:
Looks pretty good to me. One request for details in a comment. Code seems great though.

Anything else from others?

recipe/run_conda_forge_build_setup_osx
minrk: I just rephrased the prior comment, so I'm not sure. I think MACOSX_DEPLOYMENT_TARGET is set in conda_build_config.yaml by conda-smithy now that we are using cb3, so this should expect to be defined, which would make this line do nothing.

minrk: Updated the comment. It's set [in conda-forge-pinning](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/5efc55669c864b57bc9bb8b6557ee9175310d681/recipe/conda_build_config.yaml#L51) and can be overridden in recipes if need be. Setting the default here should only take effect for feedstocks still on conda-build 2.

scopatz:
Thanks @minrk!

recipe/run_conda_forge_build_setup_osx
jakirkham: Right, ok, conda-build 2 compatibility, that makes sense. Sorry wasn't thinking about that. Thanks for clarifying.

PR #35, opened 2018-11-22
Update `DTSDKName` to match our macOS SDK

jakirkham:
The `DTSDKName` was getting fixed by our old `sed` strategy along with the `MinimumSDKVersion`. We fix the `MinimumSDKVersion` already as this vital. It's less clear if the `DTSDKName` needs to be accurately reflected. Though it could show up in applications' `Info.plist`, so we might as well have it accurately reflected just in case. After this change there does not appear to be any more references to the OS SDK version other than the intended one.

Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "2.1.0" %}
+{% set version = "2.1.1" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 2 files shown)


jakirkham:
Thoughts @conda-forge/core?

scopatz:
Thanks @jakirkham!

Issue #47, opened 2019-02-12
Set CPU_COUNT correctly for Azure


jakirkham:
Currently `CPU_COUNT` is set for the previously existing CIs (i.e Travis, Circle, and AppVeyor). However to the best of my knowledge `CPU_COUNT` is likely not set correctly for Azure. Would be good to update the values set in these scripts to configure Azure correctly.

isuruf:
https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/53 has more discussion, so closing this as a duplicate of #53

PR #48, opened 2019-02-23
Activate env variables

isuruf:
cc @mariusvniekerk

### DIFF
--- a/recipe/run_conda_forge_build_setup_osx
+++ b/recipe/run_conda_forge_build_setup_osx
@@ -1,12 +1,13 @@
 #!/bin/bash
+
 # 2 core available on Travis CI OS X workers: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments
 # CPU_COUNT is passed through conda build: https://github.com/conda/conda-build/pull/1149
 export CPU_COUNT=2
 
 export PYTHONUNBUFFERED=1
 
-if [ ! -z "$CONFIG" ]; then
-    export MACOSX_DEPLOYMENT_TARGET=$(cat .ci_support/${CONFIG}.yaml | shyaml get-value MACOSX_DEPLOYMENT_TARGET.0 10.9)
+if [ -f .ci_support/${CONFIG}.yaml ]; then
+   export MACOSX_DEPLOYMENT_TARGET=$(cat .ci_support/${CONFIG}.yaml | shyaml get-value MACOSX_DEPLOYMENT_TARGET.0 10.9)
 fi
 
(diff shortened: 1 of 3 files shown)


recipe/run_conda_forge_build_setup_osx
```diff
@@ -1,11 +1,14 @@
 #!/bin/bash
-
 # 2 core available on Travis CI OS X workers: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments
 # CPU_COUNT is passed through conda build: https://github.com/conda/conda-build/pull/1149
 export CPU_COUNT=2
 
 export PYTHONUNBUFFERED=1
 
+if [ ! -z "$CONFIG" ]; then
+    export MACOSX_DEPLOYMENT_TARGET=$(cat .ci_support/${CONFIG}.yaml | shyaml get-value MACOSX_DEPLOYMENT_TARGET.0 10.9)
```
mariusvniekerk: I'd add a 
`if [ -f .ci_support/${CONFIG}.yaml ]; then` as well

recipe/run_conda_forge_build_setup_osx
```diff
@@ -1,11 +1,14 @@
 #!/bin/bash
-
 # 2 core available on Travis CI OS X workers: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments
 # CPU_COUNT is passed through conda build: https://github.com/conda/conda-build/pull/1149
 export CPU_COUNT=2
```
mariusvniekerk: Do we need to still set this?

isuruf: Yes, on the virtual machines in CI, the default values for these used by conda-build are wrong and since memory is limited, with lots of parallelism, we run out of memory.

We can change the value for azure (while keeping 2 for travis) if you think a different value is better.

isuruf: There's an open issue for azure, https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/47

mariusvniekerk: i'd probably do something like
```
export CPU_COUNT="${CPU_COUNT_OVERRIDE:-2}"
```

scopatz:
conda-verify is not available on linux-aarch64 yet.  Is this blocking?

isuruf:
No, aarch64 has been failing for a while now

mariusvniekerk:
Yeah that's not blocking and it's only for 3.6 which we're not using as our base python in any case

PR #50, opened 2019-03-07
Use strict

ocefpaf:
We should use the `strict` channel option in conda-forge since we are recommending it as the preferred way to use the channel. Also, this will act like a good test to see if `strict` is ready to be the `default` in `conda 4.7`.

The rationale behind this is that `strict` is the only way to install the GIS stack safely at the moment.

There may be problem on Windows due to the lack of `scipy`. One would expect `scipy` to be installed from `defaults` but depending on the dependencies we may get an unsolvable env. (IMO an unsolvable env is not that bad, we should focus our efforts on finding and fixing them.)

See https://github.com/conda-forge/osmnx-feedstock/pull/40#issuecomment-470362102

### DIFF
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -8,6 +8,7 @@ set PYTHONUNBUFFERED=1
 conda.exe config --set show_channel_urls true
 conda.exe config --set auto_update_conda false
 conda.exe config --set add_pip_as_python_dependency false
+conda.exe config --set channel_priority strict
 
 call setup_x64
 
--- a/.azure-pipelines/azure-pipelines-osx.yml
+++ b/.azure-pipelines/azure-pipelines-osx.yml
@@ -46,9 +46,8 @@ jobs:
       echo "Configuring conda."
 
       setup_conda_rc ./ ./recipe ./.ci_support/${CONFIG}.yaml
-      export CI=azure
-      # Overriding global run_conda_forge_build_setup_osx with local copy.
-      source recipe/run_conda_forge_build_setup_osx
+
+      source run_conda_forge_build_setup
       conda update --yes --quiet --override-channels -c conda-forge -c defaults --all
     env: {
       OSX_FORCE_SDK_DOWNLOAD: "1"
--- a/.azure-pipelines/azure-pipelines-osx.yml
+++ b/.azure-pipelines/azure-pipelines-osx.yml
@@ -46,8 +46,9 @@ jobs:
       echo "Configuring conda."
 
       setup_conda_rc ./ ./recipe ./.ci_support/${CONFIG}.yaml
-
-      source run_conda_forge_build_setup
+      export CI=azure
+      # Overriding global run_conda_forge_build_setup_osx with local copy.
+      source recipe/run_conda_forge_build_setup_osx
       conda update --yes --quiet --override-channels -c conda-forge -c defaults --all
     env: {
       OSX_FORCE_SDK_DOWNLOAD: "1"
--- a/.azure-pipelines/azure-pipelines-win.yml
+++ b/.azure-pipelines/azure-pipelines-win.yml
@@ -72,8 +72,9 @@ jobs:
 
     # Configure the VM.
     - script: |
-        run_conda_forge_build_setup
-
+        set "CI=azure"
+        # Overriding global run_conda_forge_build_setup_win with local copy.
+        recipe\run_conda_forge_build_setup_win
       displayName: conda-forge build setup
     
 
--- a/.azure-pipelines/azure-pipelines-win.yml
+++ b/.azure-pipelines/azure-pipelines-win.yml
@@ -72,9 +72,8 @@ jobs:
 
     # Configure the VM.
     - script: |
-        set "CI=azure"
-        # Overriding global run_conda_forge_build_setup_win with local copy.
-        recipe\run_conda_forge_build_setup_win
+        run_conda_forge_build_setup
+
       displayName: conda-forge build setup
     
 
--- a/.azure-pipelines/build_steps.sh
+++ b/.azure-pipelines/build_steps.sh
@@ -24,9 +24,17 @@ conda install --yes --quiet conda-forge-ci-setup=2 conda-build -c conda-forge
 # set up the condarc
 setup_conda_rc "${FEEDSTOCK_ROOT}" "${RECIPE_ROOT}" "${CONFIG_FILE}"
 
+# Overriding global run_conda_forge_build_setup_linux with local copy.
+source ${RECIPE_ROOT}/run_conda_forge_build_setup_linux
+
+# make the build number clobber
+make_build_number "${FEEDSTOCK_ROOT}" "${RECIPE_ROOT}" "${CONFIG_FILE}"
+
+conda build "${RECIPE_ROOT}" -m "${CI_SUPPORT}/${CONFIG}.yaml" \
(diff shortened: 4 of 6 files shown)


isuruf:
-1 for this. Entire R stack on Windows relies on defaults' mingw packages

ocefpaf:
> -1 for this. Entire R stack on Windows relies on defaults' mingw packages

Yep. We know. This PR is to start a discussion on how to solve that. Vendor `mingw`? Never use `strict` on Windows? Start packaging everything?

ocefpaf:
@isuruf also note that, if `strict` channel works as expected, we don't have to do anything b/c the packages that don't exist on `conda-forge` will be pulled from `defaults` anyway.

jjhelmus:
The issue with using `strict` on Windows is not the missing scipy package but rather the numpy packages in the `conda-forge` cannot meet the requirements of the `scipy` package from `defaults` since they use a different blas implementation, OpenBLAS vs MKL.

isuruf:
Thanks @jjhelmus for the info. https://github.com/conda-forge/blas-feedstock/issues/15 would fix this.

ocefpaf:
> The issue with using `strict` on Windows is not the missing scipy package but rather the numpy packages in the `conda-forge` cannot meet the requirements of the `scipy` package from `defaults` since they use a different blas implementation, OpenBLAS vs MKL.

That is what I thought but it does boil down to a missing scipy (with conda-forge deps :smile:).

BTW, I made a few tests trying to get a single package from the second channel using `strict` and virtually all of them, except the scipy/openblas windows problem, worked out just fine. One example is getting `altair`, whch does not exist in `defaults`, by adding `conda-forge` as the second channel and activating `strict`. The installation works as expected by getting everything from `defaults` but `altair`.

In light of that I do believe we should start using `strict` globally. Any issue we face are probably "fixable" by adding missing packages and or adding variants with the same deps as `defaults`.

CJ-Wright:
My understanding that the issues with windows and strict are also somewhat broken without strict, as they cause large scale movement to defaults which might not be compatible. This just makes this brokenness more obvious. Is this correct?

ocefpaf:
@isuruf would you be OK to not use strict on Windows until we fix our blas/numpy/scipy situation there but use it for Linux and macOS? It would really help those installing packages that have many dependencies, like `gdal`. (Which is currently broken for macOS thanks to mixing of packages.)

scopatz:
I think strict for mac and linux at this point in time makes a lot of sense.  If we have to hold off on windows that's OK.

ocefpaf:
> I think strict for mac and linux at this point in time makes a lot of sense. If we have to hold off on windows that's OK.

Actually, thanks to @isuruf I guess we'll be able to do it on Windows too. The problem we are hitting now is https://github.com/conda-forge/gdal-feedstock/pull/270#issuecomment-473666612

scopatz:
Should this be merged then?

isuruf:
On windows, we should wait for the BLAS migrator to finish. (It's only numpy that I have migrated manually)

ocefpaf:
> On windows, we should wait for the BLAS migrator to finish. (It's only numpy that I have migrated manually)

Thanks for that info @isuruf. I'll follow the BLAS migration closer and see if there are any PRs I can unblock.

PS: I foresee a few issues that will rise when we do this, for example, I could not start a `qt` build with strict enabled om Windows. We should identify these issues and fix them before `conda 4.7` lands.

ocefpaf:
An example where this is needed now: https://github.com/conda-forge/blas-feedstock/pull/32

borupdaniel:
Not sure if you're collecting examples where this is needed, but please feel free to delete this comment if not). Strict priority was required to successfully build/test with linux and qt5.6 in https://github.com/conda-forge/gpi-feedstock/pull/5.

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -8,6 +8,7 @@ set PYTHONUNBUFFERED=1
 conda.exe config --set show_channel_urls true
 conda.exe config --set auto_update_conda false
 conda.exe config --set add_pip_as_python_dependency false
+conda.exe config --set channel_priority strict
```
scopatz: If we got rid of this for the moment, it would probably be OK

isuruf: No, osx has issues with libcxx as well.

ocefpaf: That is "fixed" with the choice of vendorizing the old `libcxx` and may be solved for good with https://github.com/conda/conda/pull/8961.

isuruf: That's a terrible fix as I mentioned in another issue that I will revert soon. Correct fix is https://github.com/AnacondaRecipes/aggregate/pull/162

minrk:
See https://github.com/conda/conda/issues/8960 for one possible proposal to relax strict priority in the event of exact version requirements not met by the top priority channel, and related issues [for repo2docker](https://github.com/jupyter/repo2docker/issues/731 ) investigating strict priority (for frozen envs, where the broken label is our problem) and [openmpi-mpicxx](https://github.com/conda-forge/openmpi-feedstock/issues/44) where compilers coming from defaults and requiring libcxx==4.0.1 which is missing from conda-forge.

recipe/run_conda_forge_build_setup_win.bat
ocefpaf: > If we got rid of this for the moment, it would probably be OK

Can you elaborate why @scopatz? If we can list the foreseeable issues on Windows I can start working on them. (BTW, I've been using strict on Windows and Linux for a while now with more success than macOS, which should also be OK-ish now.)

isuruf:
`libcxx==4.0.1` is broken and getting it from defaults or vendoring it into conda-forge is a terrible idea. Correct fix is at https://github.com/AnacondaRecipes/aggregate/pull/162 if anybody cares to review.

ocefpaf:
> `libcxx==4.0.1` is broken and getting it from defaults or vendoring it into conda-forge is a terrible idea. Correct fix is at [AnacondaRecipes/aggregate#162](https://github.com/AnacondaRecipes/aggregate/pull/162) if anybody cares to review.


We are already doing the bad idea ;-p

I'll ping people on AR to see if we can get that merge. Thanks for the ping!

alimanfoo:
Not sure it's relevant here, but wanted to mention I'm having problems using strict channel priority and trying to install packages from bioconda and conda-forge: https://github.com/bioconda/bioconda-recipes/issues/17186

ocefpaf:
> Not sure it's relevant here, but wanted to mention I'm having problems using strict channel priority and trying to install packages from bioconda and conda-forge: [bioconda/bioconda-recipes#17186](https://github.com/bioconda/bioconda-recipes/issues/17186)

Any issues when using `strict` are "real" issue we can address. When not using `strict` you may get a solution but a broken env.

scopatz:
This has conflicts now :cry:

ocefpaf:
> This has conflicts now

Not anymore!

scopatz:
Are the blocking issues with this resolved yet?

ocefpaf:
They are not really issues but I would no merge this until more people are on board with the idea.

mwcraig (approved):
I'm on board as long as this is just affecting our build system. I still occasionally hit cases where strict is not what I want.

ocefpaf:
> I'm on board as long as this is just affecting our build system. I still occasionally hit cases where strict is not what I want.

It won't affect user options but it should make it easier to fix those cases when you cannot use `strict` that you just mentioned. Also, we need to do that before `conda 5`, when `strict` will be the `default` option.

PR #72, opened 2020-03-25
upload all built packages, except for the excluded ones

CJ-Wright:
CC: @isuruf @beckermr

### DIFF
--- a/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
+++ b/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 
 import contextlib
 import os
+import glob
 import shutil
 import subprocess
 import click
@@ -14,6 +15,7 @@ import binstar_client.errors
 from conda_build.conda_interface import subdir as conda_subdir
 from conda_build.conda_interface import get_index
 import conda_build.api
+import conda_build.config
 
 
 @contextlib.contextmanager
@@ -120,7 +122,12 @@ def upload_or_check(recipe_dir, owner, channel, variant):
     # The list of built/not skipped distributions
     built_distributions = [(m, path)
                            for m, _, _ in metas
-                           for path in conda_build.api.get_output_file_paths(m)
+                           # TODO: flip this over to .conda when that format
+                           #  is in flight
+                           for path in glob.glob(os.path.join(
+            conda_build.config.croot,
+            'noarch' if m.noarch or m.noarch_python else m.config.host_subdir,
+            '*.tar.bz2'))
                            if not m.skip()]
 
     # These are the ones that already exist on the owner channel's
--- a/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
+++ b/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
@@ -22,7 +22,7 @@ def split_pkg(pkg):
     if not pkg.endswith(".tar.bz2"):
         raise RuntimeError("Can only process packages that end in .tar.bz2")
     pkg = pkg[:-8]
-    plat, pkg_name = pkg.split("/")
+    plat, pkg_name = pkg.split(os.path.sep)
     name_ver, build = pkg_name.rsplit('-', 1)
     name, ver = name_ver.rsplit('-', 1)
     return plat, name, ver, build
--- a/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
+++ b/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
@@ -18,6 +18,16 @@ import conda_build.api
 import conda_build.config
 
 
+def split_pkg(pkg):
+    if not pkg.endswith(".tar.bz2"):
+        raise RuntimeError("Can only process packages that end in .tar.bz2")
+    pkg = pkg[:-8]
+    plat, pkg_name = pkg.split("/")
+    name_ver, build = pkg_name.rsplit('-', 1)
+    name, ver = name_ver.rsplit('-', 1)
+    return plat, name, ver, build
+
+
 @contextlib.contextmanager
 def get_temp_token(token):
     dn = tempfile.mkdtemp()
@@ -28,7 +38,7 @@ def get_temp_token(token):
     shutil.rmtree(dn)
 
 
-def built_distribution_already_exists(cli, meta, fname, owner):
+def built_distribution_already_exists(cli, name, version, fname, owner):
     """
     Checks to see whether the built recipe (aka distribution) already
     exists on the owner/user's binstar account.
@@ -39,8 +49,7 @@ def built_distribution_already_exists(cli, meta, fname, owner):
     distro_name = '{}/{}'.format(platform, basename)
 
     try:
-        dist_info = cli.distribution(owner, meta.name(), meta.version(),
-                                     distro_name)
+        dist_info = cli.distribution(owner, name, version, distro_name)
     except binstar_client.errors.NotFound:
         dist_info = {}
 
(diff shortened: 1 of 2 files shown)


recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -120,7 +122,12 @@ def upload_or_check(recipe_dir, owner, channel, variant):
     # The list of built/not skipped distributions
     built_distributions = [(m, path)
                            for m, _, _ in metas
-                           for path in conda_build.api.get_output_file_paths(m)
+                           # TODO: flip this over to .conda when that format
+                           #  is in flight
+                           for path in glob.glob(os.path.join(
+            conda_build.config.croot,
+            'noarch' if m.noarch or m.noarch_python else m.config.host_subdir,
+            '*.tar.bz2'))
```
isuruf: For each output name, you are trying to upload all files.

CJ-Wright: Can you expand on that some? I don't understand if that is what I'm trying to do and doing it incorrectly, or if that is what I should be doing and I'm not

isuruf: What this code did is for each output, calculate the build numbers and the full name(there might be many in rare cases). What the new code is doing is for each output, get the full name of all the outputs.

I think we should get rid of the conda build rendering here and just look at all the packages in `os.path.join(conda_build.config.croot, conda_build.config.subdir)` and `os.path.join(conda_build.config.croot, 'noarch')`.

One issue is that `built_distribution_already_exists` expects a conda-build meta object, but it is used for getting the name and version of the package which can be done by parsing the name as in https://github.com/conda-forge/cf-mark-broken/blob/master/mark_broken.py#L6-L13

CJ-Wright: how do you want to handle the `m.skip` logic? That depends on the rendered meta.yaml

beckermr: Why do we need skip logic at all? Shouldn't conda-build have not produced packages for things that are skipped?

CJ-Wright:
@isuruf does this implement what you had in mind?

CJ-Wright:
They don't have the needed [`.tar.bz2` extension](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/72/files#diff-2c55b5e34665621d4e425217f0f28912R120). The repodata files are either `.json` or `.json.bz2`

CJ-Wright:
I haven't checked it. I might be able to check it but not soon.

CJ-Wright:
@isuruf what is the best way to check this?

isuruf:
Copy the `upload_or_check_non_existence.py` to the recipe folder of a test feedstock and rerender. This used to work with conda-smithy, but not sure now.

isuruf:
Changing the line https://github.com/conda-forge/conda-smithy/blob/master/conda_smithy/configure_feedstock.py#L683 to `if platform == "linux"` should fix the rerendering.

CJ-Wright:
It worked! 

https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock/pull/40
https://anaconda.org/conda-forge/cf-autotick-bot-test-package/files

isuruf:
@CJ-Wright, you need a rerender

CJ-Wright:
I did a rerender, it said there was nothing to do.

CJ-Wright:
Let me try again

isuruf:
It looks like this functionality was accidentally removed from conda-smithy.

CJ-Wright:
Yep I tried again, no changes in the rerender

CJ-Wright:
what is the best path forward here?

beckermr:
I think we'd need to release a dev version of this package on a dev channel, and then hack the CI scripts in another feedstock to pull it. I think someone rewrote this to install a real python package with command-line tools and so conda-smithy cannot override them.

isuruf:
We can change the logic at https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/master/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py to use the one from the recipe directory if it exists and use it. We can merge with just that change, and then merge this PR.

beckermr:
Makes sense @isuruf. Note that smithy is using the entry point shipped with this package, so we'll want to make sure to change the logic so the entry point looks to the right thing or w/e.

isuruf:
I wasn't thinking of overriding the entry point, just the `upload_or_check_non_existence.py` file.

CJ-Wright:
My plan is to have the first line of the function see if the file is in the correct dir, if so load that file and then run the function from there.

isuruf:
Sounds good. Thanks for working on this.

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -89,6 +100,16 @@ def distribution_exists_on_channel(binstar_cli, meta, fname, owner, channel='mai
...
 def upload_or_check(recipe_dir, owner, channel, variant):
+    # If there is a dedicated upload in the recipe use that one
+    if 'upload_or_check_non_existence.py' in os.listdir(recipe_dir):
+        import importlib.util
+        spec = importlib.util.spec_from_file_location('upload_two',
+                                                      os.path.join(recipe_dir,
+                                                                   'upload_or_check_non_existence.py'))
+        mod = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(mod)
+        return mod.upload_or_check(recipe_dir, owner, channel, variant)
```
isuruf, line 111: Wouldn't this go into an endless loop?

CJ-Wright: right let me put a flag in the call.

CJ-Wright: Actually technically not, you just have to *not* have this inspection clause in the recipe version.

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -89,6 +100,16 @@ def distribution_exists_on_channel(binstar_cli, meta, fname, owner, channel='mai
 
 
 def upload_or_check(recipe_dir, owner, channel, variant):
+    # If there is a dedicated upload in the recipe use that one
+    if 'upload_or_check_non_existence.py' in os.listdir(recipe_dir):
```
isuruf, line 104:
```suggestion
    if os.path.exists(os.path.join(recipe_dir, "conda_forge_ci_setup", "upload_or_check_non_existence.py")):
```

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -89,6 +100,16 @@ def distribution_exists_on_channel(binstar_cli, meta, fname, owner, channel='mai
...
+        import importlib.util
+        spec = importlib.util.spec_from_file_location('upload_two',
+                                                      os.path.join(recipe_dir,
```
isuruf, line 107:
```suggestion
                                                      os.path.join(recipe_dir, "conda_forge_ci_setup",
```

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
isuruf: I was thinking of using the new version of this file to upload the new package, so that we don't have to test in a test feedstock. This feedstock will be its own test. That's how it was designed first.

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
CJ-Wright: Ah so the plan is to include the entire ci setup folder?

isuruf:
Can you revert back to 38e3ddd9039eeec72377d9295ba78204977add6f and merge with master?

beckermr:
It occurred to me that a more robust way to version the pinnings would be to use the time stamp of the commit. We could change the CI scripts to inject that as an environment variable and then use. 

Using the datetime at build is subject to race conditions on CI jobs and is not reproducible.

beckermr:
Cc @isuruf

isuruf:
In that case, you can add a variable like `GIT_COMMITTER_DATE` to https://github.com/conda/conda-build/blob/a563eb19bdc69552836540a80021ffe3fc417ee4/conda_build/environ.py#L159

beckermr:
Wait no. This would be the git commit hash from the repo of the source, not the feedstock. We could change the pinnings recipe to point at master, but then that would break CI. I think we would need to put this in smithy or the ci scripts here.

isuruf:
I think you can make the recipe be the source. Not sure.

beckermr:
Yup, but that would break PRs since the source would change.

beckermr:
+1 to merge this @isuruf @CJ-Wright 

it doesn't impact the change i'd like to put in specifically for versioning with the commit date

CJ-Wright:
Weird windows failures

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -14,6 +15,17 @@
...
 from conda_build.conda_interface import get_index
 import conda_build.api
+import conda_build.config
+
+
+def split_pkg(pkg):
+    if not pkg.endswith(".tar.bz2"):
+        raise RuntimeError("Can only process packages that end in .tar.bz2")
+    pkg = pkg[:-8]
+    plat, pkg_name = pkg.split("/")
```
CJ-Wright, line 25: This might not work on windows boxes. Most likely should use actual os path tools.

CJ-Wright:
@conda-forge/core can this get a merge? I'd love to get the new pinning version set up so we can merge a bunch of stuff.

PR #78, opened 2020-04-10
ENH add script to mangle homebrew

beckermr:
Checklist

### DIFF
--- a/recipe/conda_forge_ci_setup/mangle_homebrew.py
+++ b/recipe/conda_forge_ci_setup/mangle_homebrew.py
@@ -41,6 +41,7 @@ def main():
     excluded_dirs = [
         mangled_dir,
         "bin",
+        "miniconda",
     ]
 
     # move all of the stuff except miniconda
--- a/recipe/conda_forge_ci_setup/mangle_homebrew.py
+++ b/recipe/conda_forge_ci_setup/mangle_homebrew.py
@@ -43,28 +43,21 @@ def _mangele_path(pth, new_dir):
     return os.path.join(*new_parts)
 
 
-def _try_move_file(p, mangled_p):
+def _try_move_file_or_dir(p, mangled_p):
     try:
         shutil.move(p, mangled_p)
-        print("MOVED %s -> %s" % (p, mangled_p))
+        print("MOVED FILE/DIR %s -> %s" % (p, mangled_p))
     except shutil.Error:
         try:
             os.remove(p)
-            print("REMOVED %s " % p)
+            print("REMOVED FILE %s " % p)
         except Exception as e:
-            print("ERROR moving or removing %s: %s" % (p, repr(e)))
-
-
-def _try_move_dir(p, mangled_p):
-    try:
-        shutil.move(p, mangled_p)
-        print("MOVED %s -> %s" % (p, mangled_p))
-    except shutil.Error:
-        try:
-            shutil.rmtree(p, ignore_errors=True)
-            print("REMOVED %s " % p)
-        except Exception as e:
-            print("ERROR moving or removing %s: %s" % (p, repr(e)))
+            print("ERROR moving or removing FILE %s: %s" % (p, repr(e)))
+            try:
+                shutil.rmtree(p, ignore_errors=True)
+                print("REMOVED DIR %s " % p)
+            except Exception as e:
+                print("ERROR moving or removing DIR %s: %s" % (p, repr(e)))
 
 
 def main():
@@ -102,7 +95,7 @@ def main():
     for pth in KNOWN_PATHS:
         if os.path.exists(pth):
             mangled_pth = _mangele_path(pth, mangled_dir)
-            _try_move_dir(pth, mangled_pth)
+            _try_move_file_or_dir(pth, mangled_pth)
 
     # now go through the lines and move the files to a mangled path
     # if that fails, then remove them, else pass
@@ -129,7 +122,7 @@ def main():
             # and then remove
             if len(p) > 0 and os.path.exists(p) and os.path.isfile(p):
                 mangled_p = _mangele_path(p, mangled_dir)
-                _try_move_file(p, mangled_p)
+                _try_move_file_or_dir(p, mangled_p)
 
 
 if __name__ == "__main__":
--- a/recipe/conda_forge_ci_setup/mangle_homebrew.py
+++ b/recipe/conda_forge_ci_setup/mangle_homebrew.py
@@ -91,3 +91,7 @@ def main():
                         print("REMOVED %s " % p)
                     except Exception as e:
                         print("ERROR moving or removing %s: %s" % (p, repr(e)))
+
+
+if __name__ == "__main__":
+    main()
--- a/recipe/conda_forge_ci_setup/mangle_homebrew.py
+++ b/recipe/conda_forge_ci_setup/mangle_homebrew.py
@@ -39,7 +39,7 @@ def main():
     os.makedirs(mangled_dir, exist_ok=True)
 
     excluded_dirs = [
-        mangled_dir,
+        os.path.basename(mangled_dir),
         "bin",
         "miniconda",
     ]
--- a/recipe/conda_forge_ci_setup/mangle_homebrew.py
+++ b/recipe/conda_forge_ci_setup/mangle_homebrew.py
@@ -66,33 +66,34 @@ def main():
     mangled_dir = "/usr/local/mangled_homebrew_files_%s" % uuid.uuid4().hex
     os.makedirs(mangled_dir, exist_ok=True)
 
-    # we move some things we know about
-    for pth in KNOWN_PATHS:
-        if os.path.exists(pth):
+    # move all of the stuff except miniconda
+    potential_dirs = os.listdir("/usr/local")
+    for pth in potential_dirs:
+        if os.path.exists(pth) and os.path.isdir(pth) and pth not in ["bin"]:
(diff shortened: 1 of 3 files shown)


ocefpaf:
Some feedstocks need quartz from Homebrew. It would be nice to support "keeping it" in those cases.

beckermr:
@ocefpaf How does this work now? The CI scripts fully remove homebrew no matter what right now.

beckermr:
OK @conda-forge/core this script works. I tested it here: https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock/pull/43

ready for review

beckermr:
We will need a PR to smithy to enable it.

ocefpaf:
> @ocefpaf How does this work now? The CI scripts fully remove homebrew no matter what right now.

By hacking the CI config after a re-render and installing Xquartz before the removal. This script will break those feedstocks. Our original solution was already quite bad but due to the limited amount of feedstocks that needed this we never bother to do something like yum_requirements.txt.

isuruf:
@beckermr, if this is moving individual files, then it's no different from deleting.

beckermr:
🤦 

Moving directories is not really an option since that might move something we need IIUIC. I can OFC code that up, but unclear if that is a good idea. Closing.

beckermr:
@isuruf this is a filesystem dependent statement I think. I have one more idea here

isuruf:
Do you have a list of files in `/usr/local`? I think we can just delete `/usr/local`.

beckermr:
I am getting the list of files from the homebrew uninstall script using --dry-run. 

Most of them are on common paths, so I am removing those paths first. Then I go through and clean up anything else that is left.

beckermr:
This seems to work well

```
MOVED /usr/local/Caskroom -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/hee1d3c_Caskroom_hee1d3c
MOVED /usr/local/Cellar -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/h8c86b1_Cellar_h8c86b1
MOVED /usr/local/Homebrew -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/h3fbf8d_Homebrew_h3fbf8d
MOVED /usr/local/lib/gcc -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/h147faa_gcc_h147faa
MOVED /usr/local/lib/perl5 -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/h56f385_perl5_h56f385
MOVED /usr/local/include/c++ -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/hc5a118_c++_hc5a118
MOVED /usr/local/Frameworks/Python.framework -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/h948cc6_Python.framework_h948cc6
MOVED /usr/local/share/info/dir -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/hb3bbe5_dir_hb3bbe5
MOVED /usr/local/lib/node_modules/appcenter-cli/node_modules/extglob/lib/.DS_Store -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/hd1df6b_.DS_Store_hd1df6b
MOVED /usr/local/lib/node_modules/appcenter-cli/node_modules/micromatch/lib/.DS_Store -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/hfd8ffc_.DS_Store_hfd8ffc
MOVED /usr/local/lib/ruby/gems/2.6.0/gems/fastlane-2.144.0/spaceship/lib/spaceship/connect_api/.DS_Store -> /usr/local/mangled_homebrew_files_18fa6b9406364b6f97853cbf870f7ddf/h399838_.DS_Store_h399838

```

recipe/conda_forge_ci_setup/mangle_homebrew.py
```diff
@@ -0,0 +1,129 @@
...
+        else:
+            parts = [line]
+
+        for p in parts:
+            # ignore homebrew printing stuff
+            if p.startswith("==>"):
+                continue
+
+            # sometimes it does this
+            if p.startswith("Would delete "):
```
mbargull, line 116: This does not catch https://github.com/Homebrew/install/blob/db352bc12ceee10047f1da9314f624390b576697/uninstall#L244 .

beckermr: woops the space

beckermr: fixed and good catch!

mbargull:
Is this really worth it? Parsing the output of some (non-versioned) script seems really brittle.
How often do those 10 minute runs happen? IIRC 1-2 minutes are what we usually get, right?

beckermr:
I have the same thoughts here @mbargull. Plus if the problem is the homebrew script itself, then this won't help.

recipe/conda_forge_ci_setup/mangle_homebrew.py
```diff
@@ -113,8 +113,8 @@ def main():
                 continue
 
             # sometimes it does this
-            if p.startswith("Would delete "):
-                p = p[len("Would delete "):]
+            if p.startswith("Would delete"):
+                p = p[len("Would delete"):]
```
mbargull, line 117: still needs to remove the leading `:`.

beckermr:
IDK what we usually get.

mbargull:
Ca 1:30 was something I think we had, but that's just from memory (and nothing I often look at, ofc >.<).

beckermr:
Looks like 2:30 seconds here: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=137838&view=logs&jobId=4b3757f9-99c5-5891-6c0c-93d7030cdf7d&j=4b3757f9-99c5-5891-6c0c-93d7030cdf7d&t=35ac9962-a14c-5f54-e430-fcd4c09e4a5e

mbargull:
Looking at the different runs in https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=137838&view=results , it looks like the actual recipe building is more time consistent than the Homebrew removal... well, that's annoying.

beckermr:
That recipe is a shell containing nothing and not typical.

mbargull:
Sure, I just mean even `conda-build`+`conda` seem to be less erratic than Homebrew (or likely imposed I/O from it) in that case.

beckermr:
Yup it is really crazy! :/

isuruf:
If all homebrew files are under `/usr/local`, we should move everything inside `/usr/local` except `/usr/local/miniconda`.
There are other software like nodejs, powershell and other stuff installed there that we don't need.

beckermr:
OK @isuruf - we can try that. There are some caches and logs that are not in /usr/local. 

I left bin intact just in case, but I can remove that too.

isuruf:
Here's a simple bash script
https://github.com/conda-forge/_azure_pipelines_test/blob/master/azure-pipelines.yml#L42-L45

beckermr:
For sure. I thought maybe moving the code into here from smithy might help if we need to push bug fixes more quickly

isuruf:
Yes, we can have a bash script here instead of a python script.

beckermr:
I am agnostic as to bash or python, so if you'd like to use your version then, by all means, I can push it. I think the python entrypoint one is fixed now. I made a couple of stupid mistakes.

recipe/conda_forge_ci_setup/mangle_homebrew.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def main():
+    # make the mangled path
+    mangled_dir = "/usr/local/mangled_homebrew_files_%s" % uuid.uuid4().hex
+    os.makedirs(mangled_dir, exist_ok=True)
+
+    excluded_dirs = [
+        os.path.basename(mangled_dir),
+        "bin",
```
isuruf, line 43: Why keep `bin`?

beckermr: I was unsure if it had things we needed, but I can remove it.

recipe/conda_forge_ci_setup/mangle_homebrew.py
```diff
@@ -0,0 +1,61 @@
...
+    # move all of the stuff except miniconda
+    potential_dirs = os.listdir("/usr/local")
+    for _pth in potential_dirs:
+        pth = os.path.join("/usr", "local", _pth)
+        if (
+            os.path.exists(pth)
+            and os.path.isdir(pth)
+            and _pth not in excluded_dirs
+        ):
+            mangled_pth = _mangele_path(pth, mangled_dir)
```
isuruf, line 56: Let's just use `/usr/local/conda_mangled` and keep this script simple.

beckermr:
Thanks for the fixes @isuruf!

beckermr:
I'd be nice to keep the debugging statements so that ppl can see what happened if they hit issues.

isuruf:
@ocefpaf, this script will replace the script at https://github.com/conda-forge/numpy-feedstock/blob/master/.azure-pipelines/azure-pipelines-osx.yml#L33-L38 so you can still comment that out.

PR #81, opened 2020-04-16
ENH add final validation bits for CFEP-13

beckermr:
Checklist

### DIFF
--- a/recipe/conda_forge_ci_setup/build_utils.py
+++ b/recipe/conda_forge_ci_setup/build_utils.py
@@ -11,6 +11,7 @@ import sys
 import subprocess
 import click
 
+
 call = subprocess.check_call
 
 
--- a/recipe/conda_forge_ci_setup/build_utils.py
+++ b/recipe/conda_forge_ci_setup/build_utils.py
@@ -11,8 +11,6 @@ import sys
 import subprocess
 import click
 
-from conda_forge_ci_setup.upload_or_check_non_existence import retry_upload_or_check
-
 call = subprocess.check_call
 
 
@@ -136,8 +134,17 @@ def upload_package(feedstock_root, recipe_root, config_file):
                 print("Uploading to conda-forge with source channel '{}' is not allowed".format(source_channel))
                 return
 
+    upload_script_path = os.path.join(recipe_root, 'conda_forge_ci_setup', 'upload_or_check_non_existence.py')
+    if not os.path.exists(upload_script_path):
+        upload_script_path = os.path.join(os.path.dirname(__file__), 'upload_or_check_non_existence.py')
+
+    import importlib.util
+    spec = importlib.util.spec_from_file_location('upload_or_check_non_existence', upload_script_path)
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+
     for owner, channel in channels:
-        retry_upload_or_check(recipe_root, owner, channel, [config_file])
+        mod.retry_upload_or_check(recipe_root, owner, channel, [config_file])
 
 
 @click.command()
--- a/recipe/conda_forge_ci_setup/build_utils.py
+++ b/recipe/conda_forge_ci_setup/build_utils.py
@@ -11,6 +11,7 @@ import sys
 import subprocess
 import click
 
+from conda_forge_ci_setup.upload_or_check_non_existence import retry_upload_or_check
 
 call = subprocess.check_call
 
@@ -135,17 +136,8 @@ def upload_package(feedstock_root, recipe_root, config_file):
                 print("Uploading to conda-forge with source channel '{}' is not allowed".format(source_channel))
                 return
 
-    upload_script_path = os.path.join(recipe_root, 'conda_forge_ci_setup', 'upload_or_check_non_existence.py')
-    if not os.path.exists(upload_script_path):
-        upload_script_path = os.path.join(os.path.dirname(__file__), 'upload_or_check_non_existence.py')
-
-    import importlib.util
-    spec = importlib.util.spec_from_file_location('upload_or_check_non_existence', upload_script_path)
-    mod = importlib.util.module_from_spec(spec)
-    spec.loader.exec_module(mod)
-
     for owner, channel in channels:
-        mod.retry_upload_or_check(recipe_root, owner, channel, [config_file])
+        retry_upload_or_check(recipe_root, owner, channel, [config_file])
 
 
 @click.command()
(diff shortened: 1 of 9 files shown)


recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -135,7 +138,11 @@ def upload_or_check(recipe_dir, owner, channel, variant):
             for path in new_distributions:
                 upload(token_fn, path, owner, channel)
                 print('Uploaded {}'.format(path))
-        return True
+
+        if _should_validate():
+            return request_copy([b[2] for b in built_distributions], channel)
```
isuruf, line 143: Can you this inside the `with` statement? Some feedstocks have issues with deleting the temp file above.

beckermr: Changed. Note that the copy request doesn't need the binstar token in the tempfile made by the context so I don't think this should matter.

recipe/conda_forge_ci_setup/build_utils.py
```diff
@@ -135,17 +136,8 @@ def upload_package(feedstock_root, recipe_root, config_file):
                 print("Uploading to conda-forge with source channel '{}' is not allowed".format(source_channel))
                 return
 
-    upload_script_path = os.path.join(recipe_root, 'conda_forge_ci_setup', 'upload_or_check_non_existence.py')
```
isuruf, line 138: Why remove this?

recipe/conda_forge_ci_setup/build_utils.py
```diff
@@ -135,17 +136,8 @@ def upload_package(feedstock_root, recipe_root, config_file):
...
-    if not os.path.exists(upload_script_path):
-        upload_script_path = os.path.join(os.path.dirname(__file__), 'upload_or_check_non_existence.py')
-
-    import importlib.util
-    spec = importlib.util.spec_from_file_location('upload_or_check_non_existence', upload_script_path)
-    mod = importlib.util.module_from_spec(spec)
-    spec.loader.exec_module(mod)
-
```
beckermr, line 146: I removed this stuff since it is breaking the CI. We can add an option to smithy to specify a location from which to install a copy of the package.

beckermr: see here @isuruf

recipe/conda_forge_ci_setup/build_utils.py
beckermr: See my comments below. Basically, patching this one module is breaking the CI of this recipe itself. We can put an option in smithy to allow people to specify a path to this module to install the tools from or something like that.

recipe/conda_forge_ci_setup/build_utils.py
isuruf: If we remove this, how do we know that this package works?

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -17,6 +16,11 @@
 import conda_build.api
 import conda_build.config
 
+from conda_forge_ci_setup.feedstock_outputs import (
```
isuruf, line 19: Instead of importing like this, this should be done as in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/master/recipe/conda_forge_ci_setup/build_utils.py#L138-L148

recipe/conda_forge_ci_setup/build_utils.py
beckermr: We won't know it works right now because none of the validation stuff is in smithy yet. The endpoints work, but we need changes to smithy first. So it doesn't matter for that. 

I can use the import magic instead of removing this section.

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
beckermr: Right. It was one way or the other. Happy to go the other way.

isuruf: On second thought. Wouldn't `from .feedstock_output` work?

beckermr: It doesn't. I had that originally. I did not fully grep why it didn't work, but the error was something about the file not being in a module.

beckermr: To be clear, it complained about a relative import outside of a module.

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -18,6 +17,28 @@
 import conda_build.config
 
 
+def _import_feedstock_outputs_functions(recipe_root):
+    # block of code to import the feedstock_outputs module
+    feedstock_outputs_path = os.path.join(
+        recipe_root,
+        'conda_forge_ci_setup',
+        'upload_or_check_non_existence.py',
```
isuruf, line 25:
```suggestion
        'feedstock_outputs.py',
```

beckermr:
I;m testing the new smithy here, but I think it will fail.

beckermr:
@isuruf Here is the issue with how we currently do testing here. I cannot test the new entrypoint because the current package has nothing to replace.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=145605&view=logs&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=937c195f-508d-5135-dc9f-d4b5730df0f7&l=773

Any ideas?

I think a change to smithy to allow it to override the install of all of the tools instead of changing the imports file by file would be better.

isuruf:
Yes, I agree. Since the new entrypoint is not used by the other feedstocks, I'm fine with not testing that entrypoint in this PR.

isuruf:
We can also bump the version to 3.0 and test here without affecting other feedstocks. (Feedstocks pin to 2.x)

beckermr:
A major version bump is a good idea. Let's go with that.

beckermr:
OK This PR is ready to merge @isuruf @conda-forge/core. Note that per @isuruf's suggestion, we have bumped the major version to 3 here. This will let us test the package uploading functions with the new version of smithy while the rest of the org is pinned at 2.x. 

Once we are ready, we can migrate the whole org over by two changes in the conda-forge.yml and some key provisioning.

My plan is to test the uploads in the testing feedstock for the automerge stuff.

beckermr:
thanks @isuruf!

PR #82, opened 2020-04-17
BUG delete before upload, need to set the channel correctly

beckermr:
The issue we are solving with the delete is as follows. 

Suppose that the first CI run for a package uploads to cf-staging, but then the copy command given to the admin web service fails. 

Then you restart the CI service and it uploads again but this time that fails. If you then try the copy command again, it will fail because the hashes you send to it (computed from the new CI run) differ from the old CI run.

Checklist


beckermr:
@isuruf the upload on master failed due to this: 

> Failed to upload due to No module named 'ruamel'.  Trying again in 64 seconds

IDK if this is because I tried to use ruamel.yaml and ruamel_yaml or it is in the import stuff.

beckermr:
OK I removed the import lib stuff. This is all v3 anyways. I will go ahead and add code to smithy to allow ppl to install the package from a path for testing.

beckermr:
We're gonna have to push a build by hand to remove the importlib stuff. :/

beckermr:
ok @isuruf This one is ready to go. It needs to be rendered with the new version of smithy in my PR that goes with the ci setup v3 package. I have done that here and added one minor hack of including pip since the ci setup package doesn't specify that it might need it.

If you are happy with the mechanism for installing the local one to test, then we can merge this and then finish the PR on smithy too.

recipe/conda_forge_ci_setup/build_utils.py
```diff
@@ -109,43 +135,42 @@ def upload_package(feedstock_root, recipe_root, config_file):
...
-        upload_script_path = os.path.join(os.path.dirname(__file__), 'upload_or_check_non_existence.py')
-
-    import importlib.util
-    spec = importlib.util.spec_from_file_location('upload_or_check_non_existence', upload_script_path)
-    mod = importlib.util.module_from_spec(spec)
-    spec.loader.exec_module(mod)
-
     for owner, channel in channels:
-        mod.retry_upload_or_check(recipe_root, owner, channel, [config_file])
+        retry_upload_or_check(recipe_root, owner, channel, [config_file])
```
isuruf, line 173: `owner` here is still `conda-forge` right? not `cf-staging`

beckermr: I don't think so. We have to upload to cf-staging. This is the owner. It gets set dynamically depending on the `conda_forge_output_validation` parameter in the `conda-forge.yml`.

isuruf: That's not going to do anything. Target channel is set by `conda_build_config.yml` and rendered into ..ci_support files. For eg: https://github.com/conda-forge/numpy-feedstock/blob/master/.ci_support/linux_numpy1.14python3.6.____cpythonpython_implcpythontarget_platformlinux-64.yaml#L7-L8

beckermr: Frack. Thanks. I’ll fix

jakirkham:
Why are we deleting packages before uploading?

beckermr:
You have to do that for the staging channel for cfep 13

beckermr:
We have to verify the checksums and they change each time the package is built.

jakirkham:
But wouldn't this cause packages to be replaced repeatedly even for no change PRs?

beckermr:
This is for the staging stuff only, not the main channels.

recipe/conda_forge_ci_setup/build_utils.py
```diff
@@ -109,43 +133,45 @@ def upload_package(feedstock_root, recipe_root, config_file):
...
-        upload_script_path = os.path.join(os.path.dirname(__file__), 'upload_or_check_non_existence.py')
-
-    import importlib.util
-    spec = importlib.util.spec_from_file_location('upload_or_check_non_existence', upload_script_path)
-    mod = importlib.util.module_from_spec(spec)
-    spec.loader.exec_module(mod)
-
     for owner, channel in channels:
-        mod.retry_upload_or_check(recipe_root, owner, channel, [config_file])
+        if _should_validate() and owner == "conda-forge" and channel == "main":
```
isuruf, line 171: What about non main labels?

beckermr: woops yes. fixed!

beckermr:
alright @isuruf ready for another look!

beckermr:
Bump here @isuruf

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -168,18 +160,20 @@ def upload_or_check(recipe_dir, owner, channel, variant):
...
-
         with get_temp_token(cli.token) as token_fn:
             for path in new_distributions:
                 upload(token_fn, path, owner, channel)
                 print('Uploaded {}'.format(path))
 
             if _should_validate():
-                return request_copy([b[2] for b in built_distributions], channel)
+                for path in existing_distributions:
+                    delete_dist(token_fn, path, owner, channel)
```
isuruf, line 170: What's the use of this? I don't understand

beckermr: well I could be wrong but I thought anaconda uploads fail if the package already exists. 

For the staging org, we always want to upload the build we have even if it is already on the org. We have to do this so that the checksums we send match the files we then request a copy dornn

isuruf: This is problematic. There are some package with a non-noarch output and a noarch output. Noarch output is built on all platform and is basically the same for all platforms but they are not equal bit-to-bit. Whichever platform uploads first wins. Depending on who uploads first and who requests copy first, this might fail.

beckermr: I’m not following this. Can you point me to an example?

beckermr: Just to be clear, all of the stuff here happens with a platform. Also this code doesn’t remove all files for a given package, just one of them.

beckermr: I deleted my other comments since upon further thought I am following this. 

IMHO this is a bug in the package and not the code here. A given recipe should produce an output only once and upload it only once so we always know how things were built. 

In terms of a fix now, I’m not sure how to proceed. If we can make a list of the packages to fix, then we can keep the code here and fix the packages. 

Otherwise, idk how to allow for this in our current framework. Any system where the package and the thing to verify it come in two parts or are not isolated in the same transaction has this race condition.

beckermr: We could not upload noarch packages on non-linux-64 platforms. As long as every noarch package is built in at least Linux-64, then this should work.

beckermr: We could have smithy pick which  noarch package to upload and record this when rendering by having it set an env var. then we can use the env var to test if we should handle noarch packages in this code.

beckermr:
I guess you don’t want to go with the fix that works every time by listing the noarch outputs?

isuruf:
Yeah, not a huge fan of rerendering every time.

beckermr:
Sounds good to me! I’m LGTM here. If you are happy, let’s merge!

PR #87, opened 2020-04-29
Persist PATH changes

isuruf:
Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/.scripts/build_steps.sh
+++ b/.scripts/build_steps.sh
@@ -23,8 +23,7 @@ conda install --yes --quiet conda-forge-ci-setup=2 conda-build pip -c conda-forg
 
 conda uninstall --quiet --yes --force conda-forge-ci-setup
 pushd $(mktemp -d)
-cp -r ${RECIPE_ROOT}/. .
-pip install --no-deps .
+python setup.py install --single-version-externally-managed --record record.txt
 popd
 
 # set up the condarc
--- a/.scripts/build_steps.sh
+++ b/.scripts/build_steps.sh
@@ -22,9 +22,7 @@ CONDARC
 conda install --yes --quiet conda-forge-ci-setup=2 conda-build pip -c conda-forge
 
 conda uninstall --quiet --yes --force conda-forge-ci-setup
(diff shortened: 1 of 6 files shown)


isuruf:
@beckermr, any ideas why this fails on linux?

beckermr:
We are mapping directories from outside the container. Maybe the permissions are not right? I wonder if we should copy the recipe dir to a temp dir and then try again. 

If that works, note that we will need to ship a fix to smithy.

beckermr:
I think this is working. We'll have to make a PR on smithy.

beckermr:
Well I fixed the first one, but now the conda-build one is being annoying. Let me try to fix that too.

beckermr:
It a straight setuptools install works, then we should backout my changes.

beckermr:
I don't understand why pip even sees the build env version. That feels like some weird bug.

beckermr:
OK @isuruf just changing the docker mount to `rw` from `ro` fixed this without any of the other changes. Do you know why it was set at read only in the first place?

.scripts/build_steps.sh
```diff
@@ -22,7 +22,9 @@ CONDARC
 conda install --yes --quiet conda-forge-ci-setup=2 conda-build pip -c conda-forge
 
 conda uninstall --quiet --yes --force conda-forge-ci-setup
-pip install --no-deps ${RECIPE_ROOT}/.
+pushd $(mktemp -d)
+python setup.py install --single-version-externally-managed --record record.txt
```
beckermr, line 26: You would need to copy the source here.

beckermr:
I made a PR on smithy here https://github.com/conda-forge/conda-smithy/pull/1303

beckermr:
@conda-forge-admin rerender

beckermr:
OK to merge @isuruf? The change in the docker mount is in smithy so this will work out of the box once we get a smithy release.

PR #93, opened 2020-05-14
Add missing libcuda.so on 9.2

h-vetinari:
This is complementary to https://github.com/conda-forge/docker-images/pull/135; aiming at unblocking https://github.com/conda-forge/pyarrow-feedstock/pull/101 & https://github.com/conda-forge/faiss-split-feedstock/pull/1.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.0.9" %}
+{% set version = "3.0.10" %}
 
 package:
   name: {{ name|lower }}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -57,6 +57,9 @@ requirements:
     - git     # [unix]
 
 test:
+  # ensure that all variants of cuda_compiler_version are tested
+  requires:                                    # [linux and cuda_compiler_version not in (undefined, "None")]
+    - cudatoolkit {{ cuda_compiler_version }}  # [linux and cuda_compiler_version not in (undefined, "None")]
   commands:
     - if not exist "%PREFIX%\\Scripts\\run_conda_forge_build_setup.bat" exit 1    # [win]
     - test -f "${PREFIX}/bin/run_conda_forge_build_setup"                         # [unix]
--- a/recipe/run_conda_forge_build_setup_linux
+++ b/recipe/run_conda_forge_build_setup_linux
@@ -25,6 +25,26 @@ if [[ "$CONFIG" == *_pypy* ]]; then
   conda config --set channel_priority strict
 fi
 
+# the upstream image nvidia/cuda:9.2-devel-centos6 (on which linux-anvil-cuda:9.2 is based)
+# does not contain libcuda.so; it should be installed in ${CUDA_HOME}/compat-${CUDA_VER},
+# however cuda-compat-${CUDA_VER}.x86_64.rpm is only packaged for 10.x; we abuse
+# cuda-compat-10.0 for this, since the actual dependency containing libcuda for 9.2
+# (xorg-x11-drv-nvidia-libs) pulls in a huge amount of dependencies;
+# this cannot be fixed in the conda-forge linux-anvil-cuda images for licensing reasons
+# (cannot add cuda-package in our image layers), so we add it here for CI purposes only.
+if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
+  # note: $CUDA_HOME is just a symlink to /usr/local/cuda-${CUDA_VER}
+
+  # register cuda-repo with installer, cf.
+  # https://developer.download.nvidia.com/compute/cuda/repos/rhel6/x86_64/
+  curl -O https://developer.download.nvidia.com/compute/cuda/repos/rhel6/x86_64/cuda-repo-rhel6-9.2.148-1.x86_64.rpm
+  sudo yum localinstall -y cuda-repo-*.rpm
+  rm cuda-repo-*.rpm
+  # install latest cuda-compat-10-0
+  sudo yum install -y cuda-compat-10-0.x86_64 ;
+  # note: this path is added to ldconfig in linux-anvil-cuda:9.2
+  if [[ ! -f "/usr/local/cuda-10.0/compat/libcuda.so" ]]; then exit 1; fi
+fi
 
 if [ ! -z "$CONFIG" ]; then
     if [ ! -z "$CI" ]; then
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -16,6 +16,46 @@ jobs:
         CONFIG: linux_aarch64_python3.7.____cpython
         UPLOAD_PACKAGES: 'True'
         DOCKER_IMAGE: condaforge/linux-anvil-aarch64
+      linux_cuda_compiler_version10.0python3.6.____cpython:
+        CONFIG: linux_cuda_compiler_version10.0python3.6.____cpython
+        UPLOAD_PACKAGES: 'True'
+        DOCKER_IMAGE: condaforge/linux-anvil-cuda:10.0
+      linux_cuda_compiler_version10.0python3.7.____cpython:
+        CONFIG: linux_cuda_compiler_version10.0python3.7.____cpython
+        UPLOAD_PACKAGES: 'True'
+        DOCKER_IMAGE: condaforge/linux-anvil-cuda:10.0
+      linux_cuda_compiler_version10.1python3.6.____cpython:
+        CONFIG: linux_cuda_compiler_version10.1python3.6.____cpython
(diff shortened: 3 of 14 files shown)


h-vetinari:

h-vetinari:

h-vetinari:

h-vetinari:

h-vetinari:

h-vetinari:

h-vetinari:
@isuruf 
This is a necessary complement to https://github.com/conda-forge/docker-images/pull/135 - any comments?

recipe/meta.yaml
```diff
@@ -56,6 +56,9 @@ requirements:
     - git
 
 test:
+  # add requires to ensure that all variants of cuda_compiler_version are tested
+  requires:                                    # [linux and not (aarch64 or ppc64le) and cuda_compiler_version != "None"]
+    - cudatoolkit {{ cuda_compiler_version }}  # [linux and not (aarch64 or ppc64le) and cuda_compiler_version != "None"]
```
pearu, line 61: Can you eliminate `not (aarch64 or ppc64le)`? Suggestion:
```
    - cudatoolkit {{ cuda_compiler_version }}  # [cuda_compiler_version not in (undefined, "None")]
```

h-vetinari: I've had problems with `undefined` before, but let's see if it works.

h-vetinari: (and it's not like what we do now is set in stone; if the aarches gain cpu support, it would be easy to make an adaptation here.)

h-vetinari:

recipe/meta.yaml
pearu: (sure, the suggestion for just to minimize any changes to be made in future).

h-vetinari:
Ping @isuruf

recipe/run_conda_forge_build_setup_linux
```diff
@@ -25,6 +25,22 @@ if [[ "$CONFIG" == *_pypy* ]]; then
...
 fi
 
+# the upstream image nvidia/cuda:9.2-devel-centos6 (on which linux-anvil-cuda:9.2 is based)
+# does not contain libcuda.so; it should be installed in ${CUDA_HOME}/compat-${CUDA_VER},
+# however cuda-compat-${CUDA_VER}.x86_64.rpm is only packaged for 10.x; we abuse
+# cuda-compat-10.0 for this, since the actual dependency containing libcuda for 9.2
+# (xorg-x11-drv-nvidia-libs) pulls in a huge amount of dependencies;
+# this cannot be fixed in the conda-forge linux-anvil-cuda images for licensing reasons
+# (cannot add cuda-package in our image layers), so we add it here for CI purposes only.
+if [[ -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
```
isuruf, line 35:
```suggestion
if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
```

h-vetinari: Woops, thanks for the keen eye! 😅

h-vetinari:
@isuruf 
This is green and should be ready to go.

isuruf:
> error: can't create transaction lock on /var/lib/rpm/.rpm.lock (Permission denied)

h-vetinari:
OK, clearly I should stop trying to do these PRs late at night - sorry for the cumulation of oversights.

In any case, if we don't have the rights to register another rpm, then I guess we have to download directly?

h-vetinari:
OK, I give up. Even with `sudo`, I'm not getting past this
```
++ sudo mv /usr/local/cuda-10.0/compat /usr/local/cuda-9.2

We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.
```

recipe/run_conda_forge_build_setup_linux
```diff
@@ -25,6 +25,27 @@ if [[ "$CONFIG" == *_pypy* ]]; then
...
+# this cannot be fixed in the conda-forge linux-anvil-cuda images for licensing reasons
+# (cannot add cuda-package in our image layers), so we add it here for CI purposes only.
+if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
+  # see https://developer.download.nvidia.com/compute/cuda/repos/rhel6/x86_64/
+  curl -O https://developer.download.nvidia.com/compute/cuda/repos/rhel6/x86_64/cuda-repo-rhel6-9.2.148-1.x86_64.rpm && \
+  sudo yum localinstall -y cuda-repo-*.rpm && \
+  rm cuda-repo-*.rpm && \
+  sudo yum install -y cuda-compat-10-0.x86_64 && \
+  # note that $CUDA_HOME is just a symlink to /usr/local/cuda-${CUDA_VER}, and yum installs in that folder
+  sudo mv /usr/local/cuda-10.0/* /usr/local/cuda-9.2 && \
```
pearu, line 42: Could we try the following approach:
- keep /usr/local/cuda-10.0/compat as it is
- in linux-anvil-cuda/Dockerfile append /usr/local/cuda-10.0/compat to /etc/ld.so.conf.d/cuda-$CUDA_VER.conf only when CUDA_VER=9.2
?

h-vetinari: @isuruf 
What do you think about this idea? I dislike not being able to fix this correctly here, but without proper elevation - maybe we could add these two specific commands to the sudoer-list in the image?

h-vetinari: The sudoers-route works (I tested it), and should lead to IMO the cleanest solution all-around: https://github.com/conda-forge/docker-images/pull/138.

h-vetinari:
This needs to be restarted if/once https://github.com/conda-forge/docker-images/pull/138 is in.

recipe/run_conda_forge_build_setup_linux
kkraus14: I don't think this is safe to do. If a user builds a CUDA 9.2 package with this CUDA 10.0 compat `libcuda.so` this can potentially break things on a driver 396 install for example as this compat I believe is based on driver 410. I could be mistaken about this though.

@teju85 @harrism do you know if this is the case?

isuruf: The user is not building against `CUDA 10.0`. It's just loading the application using cuda compat 10.0. This is useful for checking that packages can be loaded on CI without GPUs.

kkraus14: Sorry maybe I'm unclear of the goal of moving this into the `/usr/local/cuda-9.2` folder. Is this added to `LD_LIBRARY_PATH` or something by default where it would get picked up or is this being added for someone to explicitly point to?

isuruf: Yes, it's added to ld.conf which has the almost the same effect as `LD_LIBRARY_PATH`

@h-vetinari, how about not moving that and adding `/usr/local/cuda-10.2/compat` to ld.conf in cuda 9.2 image?

kkraus14: > Yes, it's added to ld.conf which has the almost the same effect as `LD_LIBRARY_PATH`
> 
> @h-vetinari, how about not moving that and adding `/usr/local/cuda-10.2/compat` to ld.conf in cuda 9.2 image?

Won't this basically cause symbols from a newer driver library to be used during building packages which aren't backwards compatible? I believe CUDA 10.0 compat is based on driver 410, CUDA 10.2 compat is based on driver 440, but CUDA 9.2 can use driver 396 which presents possible issues from lack of backwards compatibility.

isuruf: Nope. We are using the stub library from 9.2 to build and the 10.0 compat is used only for loading

h-vetinari: > @h-vetinari, how about not moving that and adding /usr/local/cuda-10.0/compat to ld.conf in cuda 9.2 image?

This is what @pearu suggested at the top of this review thread. Fine by me.

h-vetinari: See here: https://github.com/conda-forge/docker-images/pull/139

h-vetinari:
@isuruf 
Together with the now-merged https://github.com/conda-forge/docker-images/pull/139, this PR should unblock pyarrow. LGTY?

isuruf:
@kkraus14, any objections?

kkraus14:
> @kkraus14, any objections?

I'm still not clear on how this being added to `ld.conf` can't cause problems with symbols during build, but seems like you have a solution here that I'm just not understanding 😄 so this is good to go from my perspective.

isuruf:
`ld.conf` is for runtime dynamic linker which is different to the linker `ld` used during build.

isuruf:
Thanks @h-vetinari for the PR and @kkraus14 for the review.

PR #98, opened 2020-05-26
Allow to upload privately

hadim:
Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #101, opened 2020-05-27
Expand env variables in channel names

hadim:
Checklist

### DIFF
--- a/recipe/conda_forge_ci_setup/build_utils.py
+++ b/recipe/conda_forge_ci_setup/build_utils.py
@@ -105,6 +105,7 @@ def setup_conda_rc(feedstock_root, recipe_root, config_file):
 
         call(["conda", "config", "--remove", "channels", "defaults"])
         for c in reversed(channels):
+            c = os.path.expandvars(c)
             call(["conda", "config", "--add", "channels", c])
 
         call(["conda", "config", "--set", "show_channel_urls", "true"])


hadim:
A proposal to fix https://github.com/conda-forge/conda-smithy/issues/1337

hadim:
It will allow expanding config like this:

```yaml
channels:
  sources:
    - t/$ANACONDA_TOKEN/private_channel, conda-forge, channel_2, channel_3, default
  targets:
    - private_channel main

docker:
  run_args: "-e ANACONDA_TOKEN"
```

hadim:
The following PR should fix this: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/101

recipe/conda_forge_ci_setup/build_utils.py
```diff
@@ -105,6 +105,7 @@ def setup_conda_rc(feedstock_root, recipe_root, config_file):
 
         call(["conda", "config", "--remove", "channels", "defaults"])
         for c in reversed(channels):
+            c = os.path.expandvars(c)
             call(["conda", "config", "--add", "channels", c])
```
beckermr, line 109: We’d be better off using the shell keyword argument and letting the shell do the variable expansion to avoid spilling tokens

hadim: How would you do that?

hadim: Oh the argument from `call()`! Hum... Would that work on non POSIX system?

hadim: According to https://docs.python.org/3.8/library/os.path.html#os.path.expandvars, `os.path.expandvars()` will work on Linux, OSX and Windows using `$VAR_NAME`. I don't think  `$VAR_NAME` will work on Windows with `shell=True`.

hadim: This indicates the expansion does not work on Windows (using POSIX like a variable): https://stackoverflow.com/questions/23304997/expanding-substituting-variables-in-a-popen-call#comment72190084_23305071

hadim:
Close here in favor of https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/98

recipe/conda_forge_ci_setup/build_utils.py
beckermr: Ouch. Idk how I have not hit this big before but ok.

hadim: Note that I haven't verified the statement in the SO questions...

PR #106, opened 2020-06-22
Use the latest SDK For macOS if specified

### LINKED ISSUE #67: Conda should use newer (the newest?) OSX SDK for building
Please see the discussion here:
https://github.com/conda-forge/pyside2-feedstock/issues/46#issuecomment-563911075

Conda-forge currently builds against the OSX SDK 10.9. This causes issues for example with Qt + Pyside2, which requires both Qt and python to be built with OSX SDK 10.14 or 10.15.

I understand that Conda-forge builds against SDK 10.9 to be backwards compatible to that version of OSX. However, this is also possible with building against the newest SDK (10.15), and the setting the `deployment target` to a lower OSX version, like 10.9.

Here again the details:
> > Where did you read this? That's not how how SDKs work. You build with 10.12 and it will run on 10.14, not the other way around.
> 
> See on [developer.apple.com](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/cross_development/Configuring/configuring.html#//apple_ref/doc/uid/10000163i-CH1-SW1):
> 
> It works both ways on OSX.
> 
> In the example of the link I posted, they build with base SDK 10.6 and run on 10.4 and newer. If the code doesn't use any of the newer API's (not existing in 10.6) this works out of the box. If the code wants to use new 10.5 or 10.6 API's on supported operating systems, the code must first check the OS version before using the API's. Which I guess Qt does, when they say on their [official docs](https://doc.qt.io/qt-5/macos.html) that compiling against the 10.14 SDK and then running on 10.12 works fine...
> 
> So Apple recommends always using the newest SDK, and then setting the `Deployment target` to the lowest OSX version you want to support...

psavery:
Some projects require newer versions of the OSX SDK to be used, even
though the deployment target may be older. [Qt is one of those cases](https://doc.qt.io/qt-5/macos.html),
and using an older version of the OSX SDK to build Python, for instance,
[is causing some bugs when used with PySide2](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/67).

Since Apple supports having an older deployment target while building
with a new SDK, default to the new SDK.

I am cautiously submitting this PR, without testing it, and knowing that
this might break some CI. But I am submitting it as a possible solution
to fix #67, which is causing problems for one of our projects.

Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

Fixes: #67

### DIFF
--- a/recipe/run_conda_forge_build_setup_osx
+++ b/recipe/run_conda_forge_build_setup_osx
@@ -6,6 +6,10 @@ export CPU_COUNT=2
 
 export PYTHONUNBUFFERED=1
 
+# Typically, we want to build with the lastest SDK, even though the deployment target may be older.
+# See: https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/cross_development/Configuring/configuring.html#//apple_ref/doc/uid/10000163i-CH1-SW1
+export MACOSX_SDK_VERSION=${MACOSX_SDK_VERSION:-10.15}
+
 if [ -f .ci_support/${CONFIG}.yaml ]; then
    export MACOSX_DEPLOYMENT_TARGET=$(cat .ci_support/${CONFIG}.yaml | shyaml get-value MACOSX_DEPLOYMENT_TARGET.0 10.9)
 fi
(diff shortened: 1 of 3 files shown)


psavery:

isuruf:
Please see https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/67#issuecomment-596270454 and a solution at https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/67#issuecomment-596527579

psavery:
> Please see [#67 (comment)](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/67#issuecomment-596270454) and a solution at [#67 (comment)](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/67#issuecomment-596527579)

Okay, I'll add `WITH_LATEST_OSX_SDK`.

recipe/run_conda_forge_build_setup_osx
```diff
@@ -15,15 +15,23 @@ fi
 # which should only be recipes still using conda-build 2.
 export MACOSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET:-10.9}
 
-export CONDA_BUILD_SYSROOT="$(xcode-select -p)/Platforms/MacOSX.platform/Developer/SDKs/MacOSX${MACOSX_DEPLOYMENT_TARGET}.sdk"
+if [[ "$WITH_LATEST_OSX_SDK" == "1" ]]; then
```
isuruf, line 18: This needs to be done like Line 9-11

psavery: Done.

cjh1:
I just fixed the conflicts on this, it would be great if we can get it in.

isuruf:
Thanks

PR #118, opened 2020-08-07
Look for cross compile folders too

isuruf:
Checklist

### DIFF
--- a/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
+++ b/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
@@ -15,6 +15,7 @@ from conda_build.conda_interface import subdir as conda_subdir
 from conda_build.conda_interface import get_index
 import conda_build.api
 import conda_build.config
+from conda_build.utils import DEFAULT_SUBDIRS
 
 from .feedstock_outputs import request_copy, split_pkg
 
@@ -158,7 +159,7 @@ def upload_or_check(
     allowed_dist_names = get_built_distribution_names(recipe_dir, variant)
 
     # The list of built distributions
-    paths = (
+    paths = set(
         [
             os.path.join('noarch', p)
             for p in os.listdir(os.path.join(conda_build.config.croot, 'noarch'))
@@ -169,6 +170,14 @@ def upload_or_check(
                 os.path.join(conda_build.config.croot, conda_build.config.subdir))
         ]
     )
+    DEFAULT_OS=set([subdir.split("-")[0] for subdir in DEFAULT_SUBDIRS])
+    for subdir in os.listdir(conda_build.config.croot):
+        if not subdir.split("-")[0] in DEFAULT_OS:
+            continue
+        for p in os.listdir(
+                    os.path.join(conda_build.config.croot, subdir)):
+            paths.add(os.path.join(subdir, p))
+
     built_distributions = [
         (
             split_pkg(path)[1],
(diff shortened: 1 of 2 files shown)


recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -169,6 +170,14 @@ def upload_or_check(
                 os.path.join(conda_build.config.croot, conda_build.config.subdir))
         ]
     )
+    DEFAULT_OS=set([subdir.split("-")[0] for subdir in DEFAULT_SUBDIRS])
+    for subdir in os.listdir(conda_build.config.croot):
+        if not subdir.split("-")[0] in DEFAULT_OS:
+            continue
```
beckermr, line 176: What is this if statement catching?

isuruf: There are lots of folders in `<conda-root>/conda-bld`. This removes anything that doesn't look like a subdir

isuruf:
@beckermr, I changed the logic to check for target_platform.

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -158,7 +158,7 @@ def upload_or_check(
     allowed_dist_names = get_built_distribution_names(recipe_dir, variant)
```
beckermr, line 158: I think this pulls all of the names that are built. So we can look through every subdir and make sure it is in this list.

beckermr: This check is already there: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/118/files#diff-2c55b5e34665621d4e425217f0f28912R192

So i'd just grab everything and prune it later.

isuruf: It gives use the names, but not the subdirs that they are located in

beckermr: sure but that's fine - any package with the right name can be uploaded regardless of subdir

beckermr: also the recipes are fully rendered in that function, so the correct subdir could be returned too.

beckermr:
Do we need to double check that target_subdir returns the full subdir path?

isuruf:
It shouldn't right?

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -42,7 +42,8 @@ def get_built_distribution_names(recipe_dir, variant):
     for m in skipped_distributions:
         print("{} configuration was skipped in build/skip.".format(m.name()))
 
-    return set([m.name() for m, _, _ in metas if not m.skip()])
+    subdirs = set([m.config.target_subdir for m, _, _ in metas if not m.skip()])
+    return set([m.name() for m, _, _ in metas if not m.skip()]), subdirs
```
beckermr, line 46: Sets may not be ordered. I think you need lists here.

beckermr: Instead of all combinations of names, we can just use the actual ones built.

isuruf: Why is an ordered list needed?

isuruf: Ah, I see what you are saying. I think that's a clean up for another PR.

isuruf:
Thanks for the review

PR #127, opened 2020-10-15
Remove all channels from channel list including conda-forge

isuruf:
Checklist

### DIFF
--- a/recipe/conda_forge_ci_setup/build_utils.py
+++ b/recipe/conda_forge_ci_setup/build_utils.py
@@ -103,6 +103,7 @@ def setup_conda_rc(feedstock_root, recipe_root, config_file):
             update_global_config(feedstock_root)
             channels = _global_config["channels"]["sources"]
 
+        call(["conda", "config", "--remove-key", "channels"])
         call(["conda", "config", "--remove", "channels", "defaults"])
         for c in reversed(channels):
             call(["conda", "config", "--add", "channels", c])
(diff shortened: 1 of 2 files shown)


beckermr (commented):
Why do we need this?

isuruf:
Because when we were using `miniconda`, we only had defaults and we removed defaults before adding any channels. Now we have only conda-forge, but we are removing defaults before adding any channels.

beckermr (approved):
Ah great! Thank you for explaining! LGTM!

recipe/conda_forge_ci_setup/build_utils.py
```diff
@@ -103,6 +103,7 @@ def setup_conda_rc(feedstock_root, recipe_root, config_file):
             update_global_config(feedstock_root)
             channels = _global_config["channels"]["sources"]
 
+        call(["conda", "config", "--remove", "channels", "conda-forge"])
         call(["conda", "config", "--remove", "channels", "defaults"])
         for c in reversed(channels):
             call(["conda", "config", "--add", "channels", c])
```
mbargull, line 109: Simplification:
```suggestion
        call(["conda", "config", "--remove-key", "channels"])
        for c in channels:
            call(["conda", "config", "--append", "channels", c])
```

mbargull: Ah, wait I think it'll auto-readd `defaults` if the key is empty.
This should work universally:
```suggestion
        call(["conda", "config", "--remove-key", "channels"])
        for c in channels:
            call(["conda", "config", "--append", "channels", c])
        if "defaults" not in channels:
            call(["conda", "config", "--remove", "channels", "defaults"])
```

isuruf: Thanks

PR #144, opened 2021-02-16
Add CONDA_OVERRIDE_CUDA for current CUDA version.

bdice:
Checklist

This PR adds `CONDA_OVERRIDE_CUDA` as discussed on Gitter. This will enable packages to specify `__cuda` versions as a run dependency.

I was able to test that the `shyaml` logic detects the correct values for `CONDA_OVERRIDE_CUDA`. See also: [conda docs on detected package overrides](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html#overriding-detected-packages)).

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.7.3" %}
+{% set version = "3.8.0" %}
 
 package:
   name: {{ name|lower }}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.7.2" %}
+{% set version = "3.7.3" %}
 
 package:
   name: {{ name|lower }}
--- a/recipe/run_conda_forge_build_setup_linux
+++ b/recipe/run_conda_forge_build_setup_linux
@@ -53,7 +53,7 @@ if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
 fi
 
 # Export CONDA_OVERRIDE_CUDA to allow __cuda to be detected on CI systems without GPUs
-CUDA_VERSION="$(cat ${CI_SUPPORT}/${CONFIG}.yml | shyaml get-value cuda_compiler_version.0)"
+CUDA_VERSION="$(cat ${CI_SUPPORT}/${CONFIG}.yml | shyaml get-value cuda_compiler_version None)"
 if [[ "$CUDA_VERSION" == "None" ]]; then
     CUDA_VERSION=""
 fi
--- a/recipe/run_conda_forge_build_setup_linux
+++ b/recipe/run_conda_forge_build_setup_linux
@@ -50,6 +50,10 @@ if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
   sudo yum install -y cuda-compat-10-0.$(uname -m) ;
   # note: this path is added to ldconfig in linux-anvil-cuda:9.2
   if [[ ! -f "/usr/local/cuda-10.0/compat/libcuda.so" ]]; then exit 1; fi
+
+  # Export CONDA_OVERRIDE_CUDA to allow __cuda to be detected on CI systems without GPUs
+  CUDA_VERSION="$(cat ${CI_SUPPORT}/${CONFIG}.yml | shyaml get-value cuda_compiler_version.0 None)"
+  echo "export CONDA_OVERRIDE_CUDA='${CUDA_VERSION}'" > "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
 fi
 
 if [ ! -z "$CONFIG" ]; then
--- a/recipe/run_conda_forge_build_setup_linux
+++ b/recipe/run_conda_forge_build_setup_linux
@@ -52,13 +52,6 @@ if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
   if [[ ! -f "/usr/local/cuda-10.0/compat/libcuda.so" ]]; then exit 1; fi
 fi
 
-# Export CONDA_OVERRIDE_CUDA to allow __cuda to be detected on CI systems without GPUs
-CUDA_VERSION="$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value cuda_compiler_version.0 None)"
-if [[ "$CUDA_VERSION" == "None" ]]; then
-    CUDA_VERSION=""
-fi
-export CONDA_OVERRIDE_CUDA="${CUDA_VERSION}"
-
 if [ ! -z "$CONFIG" ]; then
     if [ ! -z "$CI" ]; then
         echo "" >> ${CI_SUPPORT}/${CONFIG}.yaml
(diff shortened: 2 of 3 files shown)


recipe/run_conda_forge_build_setup_linux
```diff
@@ -50,6 +50,10 @@ if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
   sudo yum install -y cuda-compat-10-0.$(uname -m) ;
   # note: this path is added to ldconfig in linux-anvil-cuda:9.2
   if [[ ! -f "/usr/local/cuda-10.0/compat/libcuda.so" ]]; then exit 1; fi
+
+  # Export CONDA_OVERRIDE_CUDA to allow __cuda to be detected on CI systems without GPUs
+  CUDA_VERSION="$(cat ${CI_SUPPORT}/${CONFIG}.yml | shyaml get-value cuda_compiler_version.0 None)"
```
bdice, line 55: The default value should be `""` (empty string), according to the [conda docs](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html#overriding-detected-packages). I think this might fix that.
```suggestion
  CUDA_VERSION="$(cat ${CI_SUPPORT}/${CONFIG}.yml | shyaml get-value cuda_compiler_version.0 '')"
```

recipe/run_conda_forge_build_setup_linux
```diff
@@ -50,6 +50,10 @@ if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
   sudo yum install -y cuda-compat-10-0.$(uname -m) ;
   # note: this path is added to ldconfig in linux-anvil-cuda:9.2
   if [[ ! -f "/usr/local/cuda-10.0/compat/libcuda.so" ]]; then exit 1; fi
+
+  # Export CONDA_OVERRIDE_CUDA to allow __cuda to be detected on CI systems without GPUs
```
bdice, line 54: Should this block be inside or outside of the "if"?

bdice:

jakirkham:
cc @kkraus14 (for vis)

recipe/run_conda_forge_build_setup_linux
```diff
@@ -52,6 +52,13 @@ if [[ ! -z "$CUDA_HOME" && -d /usr/local/cuda-9.2 ]]; then
   if [[ ! -f "/usr/local/cuda-10.0/compat/libcuda.so" ]]; then exit 1; fi
 fi
 
+# Export CONDA_OVERRIDE_CUDA to allow __cuda to be detected on CI systems without GPUs
+CUDA_VERSION="$(cat ${CI_SUPPORT}/${CONFIG}.yml | shyaml get-value cuda_compiler_version None)"
+if [[ "$CUDA_VERSION" == "None" ]]; then
+    CUDA_VERSION=""
+fi
+echo "export CONDA_OVERRIDE_CUDA='${CUDA_VERSION}'" > "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
```
isuruf, line 60: Can you move this inside the if block?

isuruf: You also need to actually export this in this script for some CI

bdice: I moved it out of the "if" block earlier because I thought the "if" above only deals with a CUDA 9.2 workaround. It has `-d /usr/local/cuda-9.2` as a condition. Can you confirm whether that's correct? I'll add the export.

bdice:
@beckermr @jaimergp @isuruf I think this is ready for review. I would like input on how to test this, especially on Windows. The changes for Windows should be scrutinized carefully, don't trust my batch scripting to be correct just because CI passes. 😅

recipe/run_conda_forge_build_setup_linux
isuruf: No, `$CUDA_VERSION=None` is for CPU only builds, so in that case `CONDA_OVERRIDE_CUDA` should not be set at all.

kkraus14 (approved):
lgtm 😄

recipe/run_conda_forge_build_setup_linux
```diff
@@ -70,6 +77,7 @@ if [ -n "${CPU_COUNT}" ]; then
     echo "export CPU_COUNT='${CPU_COUNT}'"                  >> "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
 fi
 echo "export PYTHONUNBUFFERED='${PYTHONUNBUFFERED}'"    >> "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
+echo "export CONDA_OVERRIDE_CUDA='${CONDA_OVERRIDE_CUDA}'" >> "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
```
jaimergp, line 80: As per, Isuru's comment above, this should be only exported if CUDA_VERSION is not empty.

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -71,6 +71,7 @@ if not "%CUDA_VERSION%" == "None" (
     set "CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v%CUDA_VERSION%"
     set "PATH=%PATH%;%CUDA_PATH%\bin"
 )
+set "CONDA_OVERRIDE_CUDA=%CUDA_VERSION%"
```
jaimergp, line 74: Move this line into the if block above.

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -85,6 +86,9 @@ if not "%CUDA_PATH%" == "" (
     echo set "CUDA_PATH=%CUDA_PATH%"              >> "%CONDA_PREFIX%\etc\conda\activate.d\conda-forge-ci-setup-activate.bat"
     echo set "CUDA_HOME=%CUDA_PATH%"              >> "%CONDA_PREFIX%\etc\conda\activate.d\conda-forge-ci-setup-activate.bat"
 )
+if not "%CUDA_VERSION%" == "None" (
+    echo set "CONDA_OVERRIDE_CUDA=%CONDA_OVERRIDE_CUDA%" >> "%CONDA_PREFIX%\etc\conda\activate.d\conda-forge-ci-setup-activate.bat"
```
jaimergp, line 90: You can move this line into the if-block above so it benefits from the CUDA_PATH check (which is testing essentially for the same: did we successfully install CUDA, which implies that CUDA_VERSION is not None).

bdice: I think we have to keep the parts that write to `"%CONDA_PREFIX%\etc\conda\activate.d\conda-forge-ci-setup-activate.bat"` all together, so I don't think I can move this up any higher and eliminate the second "if CUDA" block. The first line in this section creates file with `>` and the following lines append with `>>`, so I can't know whether to create/append if I put the CUDA piece first (at the end of the first "if CUDA" block). Also we should verify the CUDA status before we start writing to that file. Does that make sense / do you have another solution in mind?

recipe/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.7.2" %}
+{% set version = "3.7.3" %}
```
isuruf, line 2:
```suggestion
{% set version = "3.8.0" %}
```

recipe/run_conda_forge_build_setup_win.bat
jaimergp: I meant the block immediately above (lines 86-87).

bdice: My mistake, I misread where the comment was pointing. Thanks, that makes complete sense now. 😆

bdice:
Thanks @isuruf @jaimergp @kkraus14 for the help and code review, it is appreciated. I have resolved all remaining comments and CI is passing now, please review/finalize/merge at your convenience. 👍

jaimergp:
LGTM!

kkraus14:
lgtm 😄

PR #153, opened 2021-04-30
Remove msmpi in System32 as well

isuruf:
Checklist

cc @leofang

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.9.3" %}
+{% set version = "3.9.4" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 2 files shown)


beckermr (approved):

leofang (approved):
LGTM. Thank you, @isuruf! This is much more robust than https://github.com/conda-forge/msmpi-feedstock/pull/11. I'll fix `msmpi` later.

leofang (commented):
On a second thought: Can we also get this line migrated to here as well?
https://github.com/conda-forge/msmpi-feedstock/blob/7fc2c26641837eb233879473ca3ee0dd27242edb/recipe/activate.bat#L10
Together this PR will do a clean removal of MS-MPI in the Windows image.

isuruf:
That's not an issue with running software. Only the DLLs in `C:\Windows\System32` are problematic. There's 100s of different software in azure that we can't possibly delete.

leofang:
The uninstaller would also clean up the registry, among other actions, to force MS-MPI to look for the correct place at runtime.

isuruf:
You mean finding smpd.exe? Looking at source code, it finds `smpd.exe` besides `msmpi.dll` first, which is the case for now. Also https://github.com/conda-forge/msmpi-feedstock/blob/master/recipe/activate.bat#L30 needs to be fixed because `LIBRARY_BIN` is not set outside of conda-build.

leofang:
Not just smpd. One of the registry keys (MSPMSProvider) points to the dll, and the key is used at initializing the communication service if you are interested in the source. Also, the pretense of the registry keys would prevent the installer from doing the right job (installing to a custom path). I'd rather keep things simple by just removing everything, which is really only one more line to add.

isuruf:
I'd like to keep the maintenance burden lower. After merging this PR, if you run into an issue, we can re-evaluate.

isuruf:
Can you open an issue in this feedstock to track this?

PR #156, opened 2021-06-30
increase pagefile size on windows

### LINKED ISSUE #155: Increase page size for azure windows agents?
After often encountering 
```
OSError: [WinError 1455] The paging file is too small for this operation to complete
```
in https://github.com/conda-forge/numpy-feedstock/pull/237 & many other pull requests (for numpy), I decided to have a quick look at what's behind that.

Stumbling over this [discussion](https://github.community/t/error-the-paging-file-is-too-small-for-this-operation-to-complete/17141) (regarding GH Actions), someone noted:
> I’ve investigated this issue and the problem is that Windows decides to create a single page file on the D:\ drive, which has only 14 GB, and by default it allocates only 1/8th of the volume size, or about 1.75 GB. This is a very small page file by modern standards and many applications will fail.

How about adding an option (or step) to the ci-setup that allows setting the page file size? It seems to be possible for GHAs - someone [published](https://github.com/actions/virtual-environments/issues/785#issuecomment-625705010) an [action](https://github.com/al-cheb/configure-pagefile-action) for it (it's MIT licensed so we could even reuse some of that code).

h-vetinari:
Fixes #155. 

As discussed in that issue, I've taken the file (& defaults) from https://github.com/al-cheb/configure-pagefile-action/blob/master/scripts/SetPageFileSize.ps1.


h-vetinari:
I see that the tests don't actually execute the `run_conda_forge_build_setup_win.bat` script - please check closely that I'm not making obvious mistakes. 😅

h-vetinari:
@isuruf, this should be ready

recipe/SetPageFileSize.ps1
```diff
@@ -0,0 +1,196 @@
...
+.SYNOPSIS
+  Configure Pagefile on Windows machine
+.NOTES
+  Author:         Aleksandr Chebotov
+.EXAMPLE
+  SetPageFileSize.ps1 -MinimumSize 4GB -MaximumSize 8GB -DiskRoot "D:"
+#>
+
+# this file taken 1:1 (with the exception of this comment) from the MIT-licensed
+# https://github.com/al-cheb/configure-pagefile-action/blob/916fa29e5d27bd4e8eef3666869afbaaee27d9eb/scripts/SetPageFileSize.ps1
```
isuruf, line 11: Can you add the copyright notice and license text here?

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -21,6 +21,11 @@ if "%CONDA_BLD_PATH%" == "" (
     set "CONDA_BLD_PATH=C:\\bld\\"
 )
 
+:: Increase pagesize, cf. https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155
+if "%CONDA_BLD_PATH%" == "C:\\bld\\" (
+    call recipe\SetPageFileSize.ps1 -MinimumSize 4GB -MaximumSize 8GB -DiskRoot "C:"
```
isuruf, line 26: Why C? C is where the building happens, so we want this to be on D

recipe/SetPageFileSize.ps1
h-vetinari: Done

recipe/run_conda_forge_build_setup_win.bat
h-vetinari: Wasn't sure if `D:`'s always available. Thanks for catching this.

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -21,6 +21,11 @@ if "%CONDA_BLD_PATH%" == "" (
     set "CONDA_BLD_PATH=C:\\bld\\"
 )
 
+:: Increase pagesize, cf. https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155
+if "%CONDA_BLD_PATH%" == "C:\\bld\\" (
```
isuruf, line 25: can you also check that `%CI%` is azure?

h-vetinari: I'm assuming this takes the value `AZURE`? Feel free to correct if wrong, I need to log off now.

recipe/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.9.4" %}
+{% set version = "3.9.5" %}
```
h-vetinari, line 2: I'm assuming this needs a new version, please revert if I'm wrong on that.

recipe/run_conda_forge_build_setup_win.bat
isuruf: It should be all lowercase

h-vetinari:
@isuruf, CI failure is spurious

h-vetinari:
@isuruf, I just looked at this again - won't we have to also copy the `SetPageFileSize.ps1` to `%SCRIPTS%`?

h-vetinari:
sorry that didn't occur to me yesterday

PR #157, opened 2021-07-01
Also copy SetPageFileSize.ps1

h-vetinari:
Fixing a very likely mistake from #156 

CC @isuruf

PR #158, opened 2021-07-11
Fix pagefile sizing; add some echos on windows

h-vetinari:
As [mentioned](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/157#issuecomment-875109296) in #157, something is still not working with resetting the pagefile size - or the 4GB are not enough for some reason.

In any case, the errors with `OSError: [WinError 1455] The paging file is too small for this operation to complete` are still occurring in the numpy feedstock (e.g. [this](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=350018&view=logs&j=171a126d-c574-5c8c-1269-ff3b989e923d&t=1183ba29-a0b5-5324-8463-2a49ace9e213) run for https://github.com/conda-forge/numpy-feedstock/pull/238), so add some echos to see what's happening.

CC @isuruf @mattip

### DIFF
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -34,7 +34,7 @@ if "%CI%" == "azure" (
     )
     if "%CONDA_BLD_PATH%" == "D:\\bld\\" (
         echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on C:
-        PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File """"%EntryPointPath%"""" -MinimumSize 8GB -MaximumSize 8GB -DiskRoot """""C:""""" '
+        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& {Start-Process PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File """"%EntryPointPath%"""" -MinimumSize 8GB -MaximumSize 8GB -DiskRoot """""C:""""" '}"
     )
 )
 
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -34,7 +34,7 @@ if "%CI%" == "azure" (
     )
     if "%CONDA_BLD_PATH%" == "D:\\bld\\" (
         echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on C:
-        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& {Start-Process PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File """"%EntryPointPath%"""" -MinimumSize """8GB""" -MaximumSize """8GB""" -DiskRoot """"C:"""" ' -NoNewWindow -Wait}"
+        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& {Start-Process PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File """"%EntryPointPath%"""" -MinimumSize ""8GB"" -MaximumSize ""8GB"" -DiskRoot """""C:""""" ' -NoNewWindow -Wait}"
     )
 )
 
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -24,10 +24,15 @@ if "%CONDA_BLD_PATH%" == "" (
 
 :: Increase pagefile size, cf. https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155
 if "%CI%" == "azure" (
+    :: use different drive than CONDA_BLD_PATH-location for pagefile
     if "%CONDA_BLD_PATH%" == "C:\\bld\\" (
-        echo Setting pagefile size to 8GB
+        echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on D:
         REM this file will be run within %SCRIPTS%, where SetPageFileSize.ps1 is also copied into, see meta.yaml
         call SetPageFileSize.ps1 -MinimumSize 8GB -MaximumSize 8GB -DiskRoot "D:"
+    )
+    if "%CONDA_BLD_PATH%" == "D:\\bld\\" (
+        echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on C:
+        call SetPageFileSize.ps1 -MinimumSize 8GB -MaximumSize 8GB -DiskRoot "C:"
     )
 )
 
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -34,7 +34,7 @@ if "%CI%" == "azure" (
     )
     if "%CONDA_BLD_PATH%" == "D:\\bld\\" (
         echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on C:
-        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& {Start-Process PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File `"%EntryPointPath%`" -MinimumSize `"8589934592`" -MaximumSize `"8589934592`" -DiskRoot `"C:`" ' -NoNewWindow -Wait}"
+        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& {Start-Process PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File """"%EntryPointPath%"""" -MinimumSize """"8589934592"""" -MaximumSize """"8589934592"""" -DiskRoot """"""""C:"""""""" ' -NoNewWindow -Wait}"
     )
 )
 
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -30,11 +30,20 @@ if "%CI%" == "azure" (
     REM use different drive than CONDA_BLD_PATH-location for pagefile
     if "%CONDA_BLD_PATH%" == "C:\\bld\\" (
         echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on D:
-        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& '%EntryPointPath%' -MinimumSize 8GB -MaximumSize 8GB -DiskRoot """"D:"""" "
+        REM Everything about this is horrible; first, we need to call PowerShell scripts differently, see:
+        REM https://blog.danskingdom.com/allow-others-to-run-your-powershell-scripts-from-a-batch-file-they-will-love-you-for-it/
+        REM Furthermore, we absolutely need to pass -DiskRoot wrapped in quotes due to the colon, but that conflicts with
+        REM the quotes necessary for the PowerShell invocation; escapes don't work, single quotes have other semantics,
+        REM and so we use the second, more involved, version of invocation from the blog (without the -Verb RunAs) that
+        REM inexplicably uses 4 quotes (and needs more options to actually see the output). This method then also manages
+        REM to break the default casting behaviour of "8GB" to UInt64, and so we pass 8 * 2^30 = 8589934592 explicitly.
+        REM The quoting of the drive letter was arrived at by trial and error (using echos in SetPageFileSize.ps1),
+        REM if you have better version (or better: an explanation), please open a PR.
+        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& {Start-Process PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File """"%EntryPointPath%"""" -MinimumSize """"8589934592"""" -MaximumSize """"8589934592"""" -DiskRoot """"""""D:"""""""""""" ' -NoNewWindow -Wait}"
     )
     if "%CONDA_BLD_PATH%" == "D:\\bld\\" (
         echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on C:
-        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& {Start-Process PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File """"%EntryPointPath%"""" -MinimumSize """"8589934592"""" -MaximumSize """"8589934592"""" -DiskRoot """"""""C:"""""""" ' -NoNewWindow -Wait}"
+        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& {Start-Process PowerShell -ArgumentList '-NoProfile -ExecutionPolicy Bypass -File """"%EntryPointPath%"""" -MinimumSize """"8589934592"""" -MaximumSize """"8589934592"""" -DiskRoot """"""""C:"""""""""""" ' -NoNewWindow -Wait}"
     )
 )
 
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -24,15 +24,15 @@ if "%CONDA_BLD_PATH%" == "" (
 
 :: Increase pagefile size, cf. https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155
 if "%CI%" == "azure" (
-    :: use different drive than CONDA_BLD_PATH-location for pagefile
+    REM use different drive than CONDA_BLD_PATH-location for pagefile
     if "%CONDA_BLD_PATH%" == "C:\\bld\\" (
         echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on D:
-        REM this file will be run within %SCRIPTS%, where SetPageFileSize.ps1 is also copied into, see meta.yaml
-        call SetPageFileSize.ps1 -MinimumSize 8GB -MaximumSize 8GB -DiskRoot "D:"
+        REM this file will be run within %SCRIPTS%, where SetPageFileSize.ps1/ExecuteSetPageFileSize.bat is also copied into, see meta.yaml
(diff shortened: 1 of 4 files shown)


h-vetinari:
Ping @isuruf, when you have a moment. :)

h-vetinari:
@isuruf, do you have any objections to merging this? We're pretty stuck with failures in the windows ci setup on the scipy-feedstock.

beckermr:
I don't care either way on this pr, but you should be able to copy this package in its entirety into the feedstock and smithy can then be configured to use it. You can then make any changes you need there for debugging. This feedstock uses that feature for testing. 

I'm happy to merge this too.

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -24,8 +25,9 @@ if "%CONDA_BLD_PATH%" == "" (
 :: Increase pagefile size, cf. https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155
 if "%CI%" == "azure" (
     if "%CONDA_BLD_PATH%" == "C:\\bld\\" (
```
isuruf, line 27: This condition is not true for numpy. You need to add another condition to check `D:\\bld\` and set the pagefile to be in `C:`

h-vetinari: Thanks, done!

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -1,3 +1,4 @@
+@echo on
```
isuruf, line 1:
```suggestion
```

h-vetinari: What's the issue with having echos in the script? The output of that batfile (in a separate step on azure, no less) is currently just the fact that it was called at all - I cannot see what would be bothersome about it, and it would help debugging (I hope) like there are currently on the scipy feedstock...

isuruf: It adds to noise in the log. The way to debug issues like these is to have an internal copy like @beckermr said and debug there.

h-vetinari: There'd be no noise on azure (because it's a separate stage), only in the raw log...

I consider myself relatively well-versed in the cf-infra, but I'd have no idea how to configure smithy to use a local copy of the ci-script (much less that this is the recommended debugging path), so I'm pretty sure this will bite other maintainers as well.

Considering how precious cf/core-resources are (I've been trying to solve this on the scipy-feedstock for almost 2 weeks with no response), I think accepting a little noise - only affecting the raw logs - to enable non-core maintainers to debug things themselves is a much better trade-off than the status quo.

mattip: The only output currently is `The system cannot find the file specified.` so any noise would be welcome to help debug the problem.

isuruf:
You can already see the problem here.

> 'SetPageFileSize.ps1' is not recognized as an internal or external command,
> operable program or batch file.

`call SetPageFileSize.ps1` is not the correct way to call a powershell script. See https://blog.danskingdom.com/allow-others-to-run-your-powershell-scripts-from-a-batch-file-they-will-love-you-for-it/

h-vetinari:
> You can already see the problem here.

Thanks for the tip on this - added a commit to try fixing it. It's still a separate issue from what's affecting the scipy feedstock (which continued working after #157 but now broke for other reasons, apparently).

h-vetinari:
PTAL @isuruf

isuruf:
This still fails in azure.

> which continued working after #157

I don't think it worked given that the if condition was never triggered.

h-vetinari:
> 
> 
> This still fails in azure.

Where can I inspect that failure in the logs? I checked before the ping already and cannot find something suspicious.

> > which continued working after #157
> 
> I don't think it worked given that the if condition was never triggered.

I'm not saying the pagefile resize worked, I meant that the whole `run_conda_forge_build_setup_win.bat` broke some time well _after_ #157. It's a separate issue that I'd like to hunt down by having better logging (which was the point of this PR, but I'm happy to fix the deficient pagefile resize en passant).

isuruf:
In https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_apis/build/builds/357385/logs/181

> 'ExecuteSetPageFileSize.bat' is not recognized as an internal or external command,
operable program or batch file.

h-vetinari:
Thanks for the pointer, I was looking in the wrong section.

> This still fails in azure.

Isn't this a chicken & egg problem? The CI for this PR is being set up with an old version of `conda-forge-ci-setup`; how can you expect a file to be there that is added to the package only a few steps later?

isuruf:
> Isn't this a chicken & egg problem? The CI for this PR is being set up with an old version of conda-forge-ci-setup; how can you expect a file to be there that is added to the package only a few steps later?

`recipe/run_conda_forge_build_setup.bat` and `recipe/ExecuteSetPageFileSize.bat` are in the same directory in the recipe and when installed, so it's just a matter of calling `ExecuteSetPageFileSize.bat` with the directory of the script `run_conda_forge_build_setup.bat`

h-vetinari:
> so it's just a matter of calling `ExecuteSetPageFileSize.bat` with the directory of the script `run_conda_forge_build_setup.bat`

Thanks, I added that and it's working now:
```
Calling: SetPageFileSize.ps1 -MinimumSize 8GB -MaximumSize 8GB -DiskRoot C:
[...]
(base) D:\a\1\s>PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& 'D:\a\1\s\recipe\SetPageFileSize.ps1' -MinimumSize 8GB -MaximumSize 8GB -DiskRoot C:" 
PageFile: 8589934592 / 8589934592 bytes for \Device\HarddiskVolume2\pagefile.sys
```

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -22,10 +23,20 @@ if "%CONDA_BLD_PATH%" == "" (
...
+set ThisScriptsDirectory=%~dp0
+set EntryPointPath=%ThisScriptsDirectory%ExecuteSetPageFileSize.bat
 if "%CI%" == "azure" (
+    REM use different drive than CONDA_BLD_PATH-location for pagefile
     if "%CONDA_BLD_PATH%" == "C:\\bld\\" (
-        REM this file will be run within %SCRIPTS%, where SetPageFileSize.ps1 is also copied into, see meta.yaml
-        call SetPageFileSize.ps1 -MinimumSize 4GB -MaximumSize 8GB -DiskRoot "D:"
+        echo CONDA_BLD_PATH=%CONDA_BLD_PATH%; Setting pagefile size to 8GB on D:
+        REM Arguments are -MinimumSize, -MaximumSize, -DiskRoot see ExecuteSetPageFileSize.bat
+        call %EntryPointPath% 8GB 8GB D:
```
isuruf, line 35: Why go with an indirection when it can just be
```suggestion
        PowerShell -NoProfile -ExecutionPolicy Bypass -Command "& '%EntryPointPath%' -MinimumSize 8GB -MaximumSize 8GB -DiskRoot D:"
```

isuruf:
There's still an error,
```
Exception calling "SetPageFileSize" with "3" argument(s): "The operation completed successfully"
At D:\a\1\s\recipe\SetPageFileSize.ps1:196 char:1
+ [Util.PageFile]::SetPageFileSize($minimumSize, $maximumSize, $diskRoo ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : Win32Exception
```

recipe/run_conda_forge_build_setup_win.bat
h-vetinari: Thanks makes sense. I just followed the blog you posted very closley because windows never ceases to amaze with trivial assumptions that end up _not_ working.

h-vetinari:
> There's still an error,

I looked again at the [reference](https://github.com/al-cheb/configure-pagefile-action/blob/master/scripts/SetPageFileSize.ps1) - the drive letter needs quotes, which clashes with the powershell invocation that also has quotes. Hopefully the quote escapes work as intended 🤞

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -22,10 +23,18 @@ if "%CONDA_BLD_PATH%" == "" (
 )
 
 :: Increase pagefile size, cf. https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155
+:: Both in the recipe and in the final package, this script is co-located with SetPageFileSize.ps1, see meta.yaml
+set ThisScriptsDirectory=%~dp0
+set EntryPointPath=%ThisScriptsDirectory%ExecuteSetPageFileSize.bat
```
isuruf, line 28:
```suggestion
set EntryPointPath=%ThisScriptsDirectory%SetPageFileSize.ps1
```

h-vetinari: 🤦
I've been travelling until today, trying to squeeze commits in here and there. Sorry for the annoying mistakes.

mattip:
At what point does it become prudent to try some other way to do this? Maybe via a python script, or compiling an executable

h-vetinari:
> At what point does it become prudent to try some other way to do this?

Haha, you're absolutely right. But right now I only have a hammer, and am trying all the ways I can dream up of hitting that nail.

If you have a workable solution, please try it with a PR. Though I'm getting closer (haven't documented all my findings yet) - now it's just about finding the right number of quotes 😆 
```

**Comment by mattip** [comment]
I have never seen a need for changing the pagefile size in CI, rather tests/compilation should be adjusted to fit in available memory. Maybe we should go back to basics and ask what feedstocks are causing this to happen.

**Comment by mattip** [comment]
As for the scipy-feedstock, isn't there any other way to figure out what is going wrong?

**Comment by h-vetinari** [comment]
I mean, there was a version before that was working _intermittently_ - sometimes it worked, sometimes it didn't. I foolishly thought it had something to do with the quoting, but it looks like all that is unnecessary, and the problem why it doesn't work occasionally is somewhere else.

It would be possible to merge this PR, it's still a strict improvement over the status quo (working half the time is better than working never), but that's up to @isuruf et al.

Otherwise, a separate PR to add the echos, or trying the work-around [mentioned](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/158#issuecomment-887864363) by @beckermr on the scipy-feedstock (though I don't know how).

**Comment by h-vetinari** [comment]
OK, this is now in the state I described above - sometimes it works - 
```
CONDA_BLD_PATH=D:\\bld\\; Setting pagefile size to 8GB on C:
Parameters:
8589934592
8589934592
C:
PageFile: 8589934592 / 8589934592 bytes for \Device\HarddiskVolume2\pagefile.sys
```
and sometimes it doesn't:
```
CONDA_BLD_PATH=D:\\bld\\; Setting pagefile size to 8GB on C:
Parameters:
8589934592
8589934592
C:
Exception calling "SetPageFileSize" with "3" argument(s): "The operation completed successfully"
At D:\a\1\s\recipe\SetPageFileSize.ps1:197 char:1
+ [Util.PageFile]::SetPageFileSize($minimumSize, $maximumSize, $diskRoo ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : Win32Exception
```

**Comment by h-vetinari** [comment]
I'm out of ideas of how to deal with this and would just like to merge this and iterate later where necessary. cc @isuruf

**Comment by mattip** [comment]
The current version seems to be hanging in the windows run of scipy-feedstock. I added `echo on` and [this is the result](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=358698&view=logs&j=e5cdccbf-4751-5a24-7406-185c9d30d021&t=d4c6ddca-7187-55a0-d44c-3d04fde5d38d)

**Comment by mattip** [comment]
Doesn't the use of the pagefile indicate a problem with the test suite? Personally I would be happier skipping a test or two in numpy instead of using the pagefile.

**Comment by h-vetinari** [comment]
> 
> 
> The current version seems to be hanging in the windows run of scipy-feedstock. I added `echo on` and [this is the result](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=358698&view=logs&j=e5cdccbf-4751-5a24-7406-185c9d30d021&t=d4c6ddca-7187-55a0-d44c-3d04fde5d38d)

Mea culpa, it seems #157 broke scipy after all. I hadn't considered that realistic because I hadn't seen other feedstocks fail similarly, but it seems that because scipy [sets](https://github.com/conda-forge/scipy-feedstock/blob/master/conda-forge.yml#L5-L9) `CONDA_BLD_PATH` to C: specifically, it is affected by the if-condition of #157 (whereas other feedstocks are not). You could try deleting that setting (though this comes from the gfortran integration PR, so probably Isuru had good reason for https://github.com/conda-forge/scipy-feedstock/commit/2c5ea1f1753b298e1a0df109cff5c6096526e19a).

**Comment by h-vetinari** [comment]
> 
> 
> Doesn't the use of the pagefile indicate a problem with the test suite? Personally I would be happier skipping a test or two in numpy instead of using the pagefile.

From the investigation in #155:

> Stumbling over this [discussion](https://github.community/t/error-the-paging-file-is-too-small-for-this-operation-to-complete/17141) (regarding GH Actions), someone noted:
> 
> > I’ve investigated this issue and the problem is that Windows decides to create a single page file on the D:\ drive, which has only 14 GB, and by default it allocates only 1/8th of the volume size, or about 1.75 GB. This is a very small page file by modern standards and many applications will fail.

I still think it would be good to set the page file size to be uniform. Intermittent failures are very annoying (and skipping tests is not a great remedy IMO - even just bisecting down to the exact set of problematic tests might be a substantial undertaking).

**Comment by mattip** [comment]
I am late to the discussion, so don't want to rehash all the good work that has been done so far. Is there a way to make the CI step to set the page file optional, document it, and have it off by default? Then at least most recipes would not need to see it.

**Comment by mattip** [comment]
Another idea: postpone this until conda-forge moves windows builds to github workflows, since then we could directly use this [github action](https://github.com/al-cheb/configure-pagefile-action) to set the pagesize?

**Comment by h-vetinari** [comment]
> Is there a way to make the CI step to set the page file optional, document it, and have it off by default? Then at least most recipes would not need to see it.

I like this idea, but not sure how I'd have to go about that. Though presumably it would be possible in a similar way like `CONDA_BLD_PATH` can be set optionally.



> 
> 
> Another idea: postpone this until conda-forge moves windows builds to github workflows, since then we could directly use this [github action](https://github.com/al-cheb/configure-pagefile-action) to set the pagesize?

I have no idea when that is going to materialise (actually, had no idea this is planned apparently), but I'd be surprised if anything happens in the next months - in that case I'd like to get it in earlier rather than later...

**Comment by mattip** [comment]
You could wrap the whole clause, from line 25 down, in an `if "%SET_PAGEFILE%" NEQ ""` condition

**Comment by xhochy** [comment]
> > Another idea: postpone this until conda-forge moves windows builds to github workflows, since then we could directly use this [github action](https://github.com/al-cheb/configure-pagefile-action) to set the pagesize?
> 
> I have no idea when that is going to materialise (actually, had no idea this is planned apparently), but I'd be surprised if anything happens in the next months - in that case I'd like to get it in earlier rather than later...

We will stick with Azure Pipelines, no plan for migrating to Github Actions. We don't have enough resources to build anything with Github Actions.

**Comment by h-vetinari** [comment]
> 
> 
> You could wrap the whole clause, from line 25 down, in an `if "%SET_PAGEFILE%" NEQ ""` condition

I like this idea - WDYT @isuruf?

**Comment by h-vetinari** [comment]
I added a commit for the guard now; I can also open a PR that just removes the pagefile stuff completely (to unblock scipy), but I still think it's worth keeping for the problems on the numpy feedstock, and this PR would be ready.

File: recipe/run_conda_forge_build_setup_win.bat [hunk 5]
```diff
@@ -22,10 +22,24 @@ if "%CONDA_BLD_PATH%" == "" (
...
 :: Increase pagefile size, cf. https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/155
-if "%CI%" == "azure" (
-    if "%CONDA_BLD_PATH%" == "C:\\bld\\" (
-        REM this file will be run within %SCRIPTS%, where SetPageFileSize.ps1 is also copied into, see meta.yaml
-        call SetPageFileSize.ps1 -MinimumSize 4GB -MaximumSize 8GB -DiskRoot "D:"
+:: Both in the recipe and in the final package, this script is co-located with SetPageFileSize.ps1, see meta.yaml
+set ThisScriptsDirectory=%~dp0
+set EntryPointPath=%ThisScriptsDirectory%SetPageFileSize.ps1
+:: Only run if SET_PAGEFILE is set; EntryPointPath needs to be set before if not using delayed execution.
+if "%SET_PAGEFILE%" NEQ "" (
```
mattip, line 29: Reorder the lines to scope environment variables and simplify comment (the "if" is self-explanatory, I don't understand the comment about EntryPointPath so I left it untouched: what is delayed excecution?)
```suggestion
if "%SET_PAGEFILE%" NEQ "" (
    :: Both in the recipe and in the final package, this script is co-located with 
    :: SetPageFileSize.ps1, see meta.yaml
    set ThisScriptsDirectory=%~dp0
    set EntryPointPath=%ThisScriptsDirectory%SetPageFileSize.ps1
    :: EntryPointPath needs to be set before if not using delayed execution.
```

h-vetinari: If the variables are set within the if-condition, they are not available for use within the script (yeah, windows...), unless delayed ~execution~ expansion is used: https://ss64.com/nt/delayedexpansion.html

That's potentially a larger change though that I didn't want to touch, and setting the variables before the if-condition doesn't hurt anyone, IMO.

mattip: Ahh, ok. Never mind then.

mattip:
LGTM, what else is needed to merge this?

h-vetinari:
@isuruf, could you please give a go/no-go here?

h-vetinari:
Thanks for your patience on this one, everyone, and apologies again for getting it wrong the first time. Fingers crossed this will now work better (I might later raise an issue on https://github.com/al-cheb/configure-pagefile-action - where the powershell script is from - to see what they think about the `Exception calling "SetPageFileSize" with "3" argument(s): "The operation completed successfully"`)

PR #182, opened 2022-01-30
ENH support .conda format

beckermr:
Checklist

This PR adds support for .conda artifacts.


beckermr:
@conda-forge-admin  rerender

beckermr:
@conda-forge/core Is this recipe supposed to generate ~76 variants? I guess yes, it just seems like a lot.

beckermr:
I guess we could build some downstream CUDA tests for all of the variants and list them here. We'd have to update the list here manually which kind of stinks. I think the right answer here is some sort of intrinsic testing loop over variants within conda build, but who has the time!

Moving on.

beckermr:
@conda-forge-admin rerender

beckermr:
@conda-forge-admin rerender

recipe/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.18.3" %}
+{% set version = "3.18.4" %}
```
beckermr, line 2:
```suggestion
{% set version = "3.18.5" %}
```

beckermr:
@conda-forge-admin rerender

beckermr:
@conda-forge-admin rerender

xhochy (approved):

beckermr:
Thanks @xhochy. This needs to be tested before it is merged.

xhochy:
Code was looking but I was under the impression that this would simply be for preparation and we could not really test this yet?

beckermr:
We can copy the package wholesale to our test feedstock and it will get run live there. Not going to do that today but will do before we merge.

beckermr:
This one is being tested here: https://github.com/conda-forge/cf-autotick-bot-test-package-feedstock/pull/439

jakirkham:
Should we go ahead and add that key to feedstocks, but set it to `false`?

beckermr:
You would want to set it to `None`, but no we should not.

beckermr:
Or at least we don't gain anything by setting it to `None` now since old versions of the ci setup package simply ignore it.

jakirkham:
Sorry `None`. Oh what is the issue with setting it?

beckermr:
I'm not following sorry. The possible values are

```
conda_pkg_format: None  # .tar.bz2
conda_pkg_format: "2"   # .conda
```

The code here does

```
def maybe_use_dot_conda(feedstock_root):
    """Maybe set the .condarc to use .conda files."""
    if os.path.exists(os.path.join(feedstock_root, "conda-forge.yml")):
        with open(os.path.join(feedstock_root, "conda-forge.yml")) as f:
            repo_config = safe_load(f)
        pkg_format = repo_config.get("conda_pkg_format", None)
        if pkg_format is not None:
            try:
                call([
                    "conda", "config", "--env", "--set",
                    "conda_build.pkg_format", str(pkg_format)
                ])
            except subprocess.CalledProcessError:
                pass
```

Once we want to move the default, we change 

```
pkg_format = repo_config.get("conda_pkg_format", None)
```

to 

```
pkg_format = repo_config.get("conda_pkg_format", "2")
```

So the current default of tar.bz2 is hard coded into our scripts until we change it. 

Does that answer your question?

recipe/conda_forge_ci_setup/build_utils.py
```diff
@@ -107,16 +107,34 @@ def fail_if_travis_not_allowed_for_arch(config_file, feedstock_root):
...
+            repo_config = safe_load(f)
+        pkg_format = repo_config.get("conda_pkg_format", None)
+        if pkg_format is not None:
+            try:
+                call([
+                    "conda", "config", "--env", "--set",
+                    "conda_build.pkg_format", str(pkg_format)
+                ])
+            except subprocess.CalledProcessError:
+                pass
```
mbargull, line 123: Why is it allowed to pass?

beckermr: Well, do we want to stop a feedstock from building due to an error here?

beckermr: For the record, we might. I wasn't sure on the first pass though since the old artifact style still works.

mbargull: Yes?! Why not? We'd want to build with the expected config (and be informed if it doesn't work), don't we?

mbargull: > For the record, we might. I wasn't sure on the first pass though since the old artifact style still works.

Ah, if this is meant as a temporary thing until we are sure all things are working, then that's fine (needs a `TODO`/`FIXME` comment then though ;P ).

beckermr, line 123:
```suggestion
            call([
                "conda", "config", "--env", "--set",
                "conda_build.pkg_format", str(pkg_format)
            ])
```

beckermr: fixed! thanks for the review!

beckermr:
@conda-forge/core any more comments here?

PR #190, opened 2022-02-20
MNT: rerender

### LINKED ISSUE #189: @conda-forge-admin rerender
- [ ] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

Issue:

<br/>
Environment (<code>conda list</code>):
<details>

```
$ conda list

```
</details>

<br/>
Details about  <code>conda</code> and system ( <code>conda info</code> ):
<details>

```
$ conda info

```
</details>

conda-forge-linter:
Hi! This is the friendly automated conda-forge-webservice.

I've rerendered the recipe as instructed in #189.

Here's a checklist to do before merging.
- [ ] Bump the build number if needed.

Fixes #189


beckermr:
@conda-forge-admin rerender

conda-forge.yml
```diff
@@ -10,3 +10,4 @@ test_on_native_only: true
 github:
   branch_name: main
   tooling_branch_name: main
+build_with_mambabuild: False
```
beckermr, line 13: Can be removed once https://github.com/conda-forge/conda-smithy/pull/1594 is merged and released.

beckermr:
ouch:  `- nvcc_linux-64=None`

recipe/meta.yaml
```diff
@@ -37,9 +37,15 @@ build:
...
   skip: True  # [py2k]
+  ignore_run_exports_from:
+    - {{ compiler('cuda') }}              # [cuda_compiler_version != None]
+    - {{ compiler('c') }}                 # [cuda_compiler_version != None]
 
 requirements:
   build:
+     # ensure that all variants of cuda_compiler_version are tested
+    - {{ compiler('cuda') }}              # [cuda_compiler_version != None]
+    - {{ compiler('c') }}                 # [cuda_compiler_version != None]
```
beckermr, line 48:
```suggestion
    - {{ compiler('cuda') }}              # [cuda_compiler_version != "None"]
    - {{ compiler('c') }}                 # [cuda_compiler_version != "None"]
```

Maybe this?

recipe/meta.yaml
```diff
@@ -37,9 +37,15 @@ build:
     - mangle_homebrew = conda_forge_ci_setup.mangle_homebrew:main
     - validate_recipe_outputs = conda_forge_ci_setup.feedstock_outputs:main
   skip: True  # [py2k]
+  ignore_run_exports_from:
+    - {{ compiler('cuda') }}              # [cuda_compiler_version != None]
+    - {{ compiler('c') }}                 # [cuda_compiler_version != None]
```
beckermr, line 42:
```suggestion
    - {{ compiler('cuda') }}              # [cuda_compiler_version != "None"]
    - {{ compiler('c') }}                 # [cuda_compiler_version != "None"]
```

jakirkham:
cc @jaimergp

jaimergp:
I am missing a bit of context here. I guess this is to reduce the huge build matrix we had due to the CUDA versions and sync with `conda-forge-pinning`?

beckermr:
@conda-forge-admin rerender

conda-forge.yml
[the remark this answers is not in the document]
beckermr: Thanks for this! We'll need to wait another hour or so for the CDN.

hmaarrfk:
Is this going to create many packages with the same build number, but different build strings (due to the introduction of cuda).

I think this is going to cause the solvers some real headaches with users flip flopping between equivalent releases.

beckermr:
We could/should bump the version to prevent this maybe. 

The "solver headaches" ship has largely sailed now. The repodata is so big that optimizing exactly what builds we push in an effort to avoid solver issues is not going to yield a bunch of nice results IMHO.

hmaarrfk:
From a user experience point of view, I often find myself (and people I explain to) confused by packages like:

```
removing version 1.23.45 build 0 hash 123456
installing version 1.23.45 build 0 hash 7890ABC
```
The solver in this case has no way to prefer one over the other, and I've found it to often flip flop between the two when I type:
```
mamba update --all --yes
```

I would like to avoid this flip flopping if we can. But maybe you are right and better testing is more important.

beckermr:
100% agreed! The version bump will avoid it. Given the changes here, I think it makes sense.

beckermr:
@conda-forge-admin rerender

hmaarrfk:
Is conda-forge-ci-setup dependent on `compilers('cuda')` that seems strange to me.

beckermr:
Yes. We install the windows cuda packages on the fly.

hmaarrfk:
Got it. Then why not add cuda to the build string? Isnt that what we do for other packages?

beckermr:
Ahhhhh. So the cuda variants generate tests of the package during CI with cuda. At runtime, the cuda variant doesn't need to be matched AFAIK. We could prioritize the cuda None package by build number maybe to prevent the issue above. I think I've been not understanding you.

hmaarrfk:
Right. I'm not sure what the best solution is. I would prefer to revert to the old way.

To test against different cuda variants, I would use a selector based on the python version.

* python 3.9 tests with cuda None
* python 3.10 tests with cuda 11.2
* python 3.8 tests with .....

That limits the number of outputs, but yet does a good job testing against different cuda variations.

beckermr:
> I would prefer to revert to the old way.

You mean one build per platform and python version?

hmaarrfk:
> > I would prefer to revert to the old way.
> 
> You mean one build per platform and python version?

Yes.

beckermr:
I bet the build number approach would be easier to maintain and will get us to the same result. If you want to make a separate PR for the other way, that is fine. 

The purpose of this PR was only to make sure rerendering worked for the switch from master to main and to generate some builds on main. 

Thus I think we should defer this improvement to another PR.

beckermr:
Any other comments @hmaarrfk @conda-forge/core or can we merge?

hmaarrfk:
good for me.

beckermr:
FYI I opened an issue (#191).

beckermr:
@conda-forge-admin rerender

beckermr:
I'm testing the build number bump idea here to make sure it works. I can remove if we want to merge but I am feeling less comfortable merging than I was before.

recipe/meta.yaml
```diff
@@ -1,5 +1,10 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.18.2" %}
+{% set version = "3.18.3" %}
+{% set build = 0 %}
+{% if cuda_compiler_version in (undefined, None, "None") %}
```
beckermr, line 4:
```suggestion
{% if cuda_compiler_version in (None, "None") %}
```

beckermr:
```suggestion
{% set cuda_compiler_version = cuda_compiler_version or "None" %}
{% if cuda_compiler_version == "None" %}
```

recipe/meta.yaml
```diff
@@ -1,5 +1,10 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.18.2" %}
+{% set version = "3.18.3" %}
+{% set build = 0 %}
+{% if cuda_compiler_version in (None, "None") %}
```
beckermr, line 4:
```suggestion
{% set cuda_compiler_version = cuda_compiler_version or "None" %}
{% if cuda_compiler_version == "None" %}
```

beckermr:
@conda-forge-admin rerender

PR #202, opened 2022-05-30
use latest compatible conda

wolfv:
@isuruf please note that this doesn't fix anything as we have the same constraint in `boa` which is actually triggering the issue. However, this will at least force the latest compatible conda version to be used.


wolfv:
@isuruf I don't think this is technically "needed" anymore. What do you think?

PR #209, opened 2022-10-31
cross compile cuda support

isuruf:
Checklist

cc @jakirkham, @kkraus14, @h-vetinari

### DIFF
--- a/recipe/cross_compile_support.sh
+++ b/recipe/cross_compile_support.sh
@@ -2,9 +2,11 @@ BUILD_PLATFORM=$(conda info --json | jq -r .platform)
 
 if [ -f ${CI_SUPPORT}/${CONFIG}.yaml ]; then
     HOST_PLATFORM=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value target_platform.0 ${BUILD_PLATFORM})
+    CUDA_COMPILER_VERSION=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value cuda_compiler_version.0 None)
 fi
 
 HOST_PLATFORM=${HOST_PLATFORM:-${BUILD_PLATFORM}}
+CUDA_COMPILER_VERSION=${CUDA_COMPILER_VERSION:-None}
 
 if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
     echo "export CONDA_BUILD_CROSS_COMPILATION=1"                 >> "${CONDA_PREFIX}/etc/conda/activate.d/conda-forge-ci-setup-activate.sh"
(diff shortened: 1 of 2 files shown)


recipe/cross_compile_support.sh
```diff
@@ -15,19 +17,66 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+                        libcurand-devel-11-2-10.2.3.152-1 \
+                        libcusolver-11-2-11.1.0.152-1 \
+                        libcusolver-devel-11-2-11.1.0.152-1 \
+                        libcusparse-11-2-11.4.1.1152-1 \
+                        libcusparse-devel-11-2-11.4.1.1152-1 \
+                        libnpp-11-2-11.2.1.68-1 \
+                        libnpp-devel-11-2-11.2.1.68-1 \
+                        libnvjpeg-11-2-11.4.0.152-1 \
+                        libnvjpeg-devel-11-2-11.4.0.152-1 ; do
+                    curl -L -O https://developer.download.nvidia.com/compute/cuda/repos/rhel8/${CUDA_HOST_PLATFORM_ARCH}/${f}.${HOST_PLATFORM_ARCH}.rpm
```
kkraus14, line 69: Is there any potential issues with glibc versioning with using rhel8 cross compilation libraries in a cos7 (or even cos6) sysroot?

isuruf: No. Only constraint is that we have to match the distro we use for native compiling which is centos 8 for cuda>=11.2 for ppc64le and aarch64.

h-vetinari:
Thanks @isuruf, that's great!

I guess one thing we have to think about is how to square this with the zip (in the current migrator) that pulls in a specific docker-image. On the arrow-feedstock I worked around this with a commit such as https://github.com/conda-forge/arrow-cpp-feedstock/pull/859/commits/61be7392b443f75ee4af92fc4cce5423ddcafbd7, but I think a more proper solution would be https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3624

recipe/cross_compile_support.sh
```diff
@@ -15,19 +19,71 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+
+        if [[ "${CUDA_COMPILER_VERSION}" == "11.2" ]] && [[ "${CDT_NAME}" == "cos8" || "${CDT_NAME}" == "cos7" ]]; then
+            # We use cdt_name=cos7 for rhel8 based nvcc till we figure out
+            # a stable cos8 replacement.
+            EXTRACT_DIR=$(mktemp -d)
+            pushd ${EXTRACT_DIR}
+                if [[ "${HOST_PLATFORM_ARCH}" == "aarch64" ]]; then
+                    CUDA_HOST_PLATFORM_ARCH=sbsa
+                else
+                    CUDA_HOST_PLATFORM_ARCH=${HOST_PLATFORM_ARCH}
```
h-vetinari, line 47: IIUC, we should error out here in the PPC-case? (seeing that we're already in `[[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]] && [[ "${CUDA_COMPILER_VERSION}" == "11.2" ]] && ...` at this point)

isuruf: It turns out that ppc64le is also supported.

h-vetinari:
@isuruf, lemme know if there's anything I can do to help with this PR

isuruf:
See Keith's comment above

h-vetinari:
> See Keith's comment above

I gave it a shot in #210, but I'm not sure if it's correct; I let it run during the test phase and the CI is green, so it doesn't seem too far off.

Worth noting: the manifest posted by Keith only contains one devel package (`cuda-nvml-devel`), whereas this PR so far has:
```
                        cuda-cudart-devel-11-2-11.2.72-1\
                        cuda-driver-devel-11-2-11.2.152-1 \
                        cuda-nvml-devel-11-2-11.2.67-1 \
                        cuda-nvrtc-devel-11-2-11.2.152-1 \
                        libcublas-devel-11-2-11.3.1.68-1 \
                        libcufft-devel-11-2-10.4.1.152-1 \
                        libcurand-devel-11-2-10.2.3.152-1 \
                        libcusolver-devel-11-2-11.1.0.152-1 \
                        libcusparse-devel-11-2-11.4.1.1152-1 \
                        libnpp-devel-11-2-11.2.1.68-1 \
                        libnvjpeg-devel-11-2-11.4.0.152-1
```

If we need to add those devel packages (most likely necessary to have the headers?!), then that needs more massaging, but shouldn't be too hard.

PR #210, opened 2022-11-04
cross compile cuda support (cnt'd)

h-vetinari:
New PR since I cannot commit into #209

Closes #209


recipe/cross_compile_support.sh
```diff
@@ -15,19 +19,52 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+            # We use cdt_name=cos7 for rhel8 based nvcc till we figure out
+            # a stable cos8 replacement.
+            EXTRACT_DIR=$(mktemp -d)
+            pushd ${EXTRACT_DIR}
+                if [[ "${HOST_PLATFORM_ARCH}" == "aarch64" ]]; then
+                    CUDA_HOST_PLATFORM_ARCH=sbsa
+                else
+                    CUDA_HOST_PLATFORM_ARCH=${HOST_PLATFORM_ARCH}
+                fi
+                curl -L https://developer.download.nvidia.com/compute/cuda/repos/rhel8/sbsa/version_11.2.2.json > versions.json
```
kkraus14, line 49: Should we use `CUDA_HOST_PLATFORM_ARCH` instead of hardcoding to `sbsa` for the URL here? I imagine the versions are consistent but I'd rather be safe than sorry.

recipe/cross_compile_support.sh
```diff
@@ -15,19 +19,52 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+                done
+                mv ./usr/local/cuda-${CUDA_COMPILER_VERSION}/targets/${CUDA_HOST_PLATFORM_ARCH}-linux ${CUDA_HOME}/targets/${CUDA_HOST_PLATFORM_ARCH}-linux
+            popd
+            rm -rf ${EXTRACT_DIR}
+        elif [[ "${CUDA_COMPILER_VERSION}" == "11.2" ]]; then
+	    echo "cross compiling with cuda == 11.2 and cdt != cos7/8 not supported yet"
+	    exit 1
+        elif [[ "${CUDA_COMPILER_VERSION}" != "None" ]]; then
+	    # FIXME: can use anaconda.org/nvidia packages to get the includes and libs
+	    # for cuda >=11.3.
```
kkraus14, line 65: Do we care about supporting cross compilation for CUDA 11.0 or 11.1? My take is probably not, but figure I'd ask.

h-vetinari: I'd say only 11.2+

recipe/cross_compile_support.sh
h-vetinari: I think that should be possible.

h-vetinari:
Thanks for the quick review @kkraus14! Do you know whether we need the other [devel](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/209#issuecomment-1302890213) packages?

kkraus14:
> Thanks for the quick review @kkraus14! Do you know whether we need the other [devel](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/209#issuecomment-1302890213) packages?

Yes we'll need those. Those contain the headers, the generic unversioned symlinked shared library, the static libraries that we may want specifically for CUDA because there's some implications IIRC, and the pkgconfig files.

recipe/cross_compile_support.sh
```diff
@@ -15,19 +22,79 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+                else
+                    CUDA_HOST_PLATFORM_ARCH=${HOST_PLATFORM_ARCH}
+                fi
+                # download manifest for latest 11.2.x patch version
+                curl -L https://developer.download.nvidia.com/compute/cuda/repos/rhel8/${CUDA_HOST_PLATFORM_ARCH}/version_11.2.2.json > versions.json
+                # packages for which we also need to install the -devel version
+                # (names need "_" not "-" to match spelling in manifest)
+                declare -a DEVELS=(
+                    "cuda_cudart"
+                    "cuda_driver"
```
kkraus14, line 58: I believe you need to use the `cuda_cudart` version for `cuda_driver` since it doesn't have its own key.

h-vetinari:
Argh, the test here cannot work because at test time, the "BUILD_PLATFORM" is already emulated.  

> Yes we'll need those.

Added in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210/commits/91610af356278676b9ee4dabc5afbc14600e1714; not sure if my scripting is worth a damn, I'm not experienced with `jq`, but I tested it in https://jqplay.org/ at least, and of course bash errors are a real possibility too. 😅

recipe/cross_compile_support.sh
h-vetinari: Done in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210/commits/a7d3bda1722c2805a05f801f96fe4f06f673de07

recipe/cross_compile_support.sh
```diff
@@ -15,19 +22,79 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+                    "cuda_nvml"
+                    "cuda_nvrtc"
+                    "libcublas"
+                    "libcufft"
+                    "libcurand"
+                    "libcusolver"
+                    "libcusparse"
+                    "libnpp"
+                    "libnvjpeg"
+                )
```
kkraus14, line 68: Need to add: `nvidia_driver`, `nvidia_fabric_manager`, `nvidia_libXNVCtrl`. All three should use the version from the `nvidia_driver` key

recipe/cross_compile_support.sh
```diff
@@ -15,19 +22,79 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+                echo "jq-query: $DEVELQUERY"
+                # update versions.json with devel packages
+                jq '${DEVELQUERY}' versions.json > versions.json
+                # sanity check
+                cat versions.json
+                # read names & versions for necessary RPMs from manifest; download & install them
+                jq 'keys[] as $k | "\($k)-11-2-\(.[$k] | .version)-1"' versions.json | while read d; do
+                    # normalize "_" -> "-"; also adapt "_dev" -> "-devel" (for cuda_nvml_dev)
+                    fn="$(echo $d | sed 's/_/-/g' | sed 's/-dev/-devel/g').${HOST_PLATFORM_ARCH}.rpm"
+                    echo "Downloading & installing: $fn"
```
kkraus14, line 83: I think we want to filter down to the `-dev` and `-devel` packages here. I think that's what you're doing here, but my bash reading skills are failing me at the moment.

The runtime packages will generally be non-functional without a GPU anyway.

h-vetinari: To get an idea what I'm doing here, copy [payload](https://developer.download.nvidia.com/compute/cuda/repos/rhel8/sbsa/version_11.2.2.json) into https://jqplay.org/ and then add `keys[] as $k | "\($k)-11-2-\(.[$k] | .version)-1"` as the filter. The RHS then gives you the list of things I'm looping over.

To get from the contents of the manifest to the rpm-names, I need to replace "_" with "-" (in addition to the string interpolation in the jq query), and I'm replacing the `cuda_nvml_dev` with `cuda-nvml-devel`.

That output in the playground is without the extra devel-files being generated, but you can get an idea of what it's trying to do by replacing the filter with

```
. += {"cuda_devel": {"version": .cuda.version}} | keys_unsorted[] as $k | "\($k)-11-2-\(.[$k] | .version)-1"
```

recipe/cross_compile_support.sh
h-vetinari: OK, this and the `cuda_driver` thing make it sound like we need to do a full mapping here. Currently it was just `<x> -> `<x>_devel`.


You're saying I need to change this to:
```
    # new key...                ...uses version from
    cuda_driver_devel       ->  cuda_cudart       # special
    cuda_nvrtc_devel        ->  cuda_nvrtc
    cuda_nvml_devel         ->  cuda_nvml_dev     # special
    libcublas_devel         ->  libcublas
    libcufft_devel          ->  libcufft
    libcurand_devel         ->  libcurand
    libcusolver_devel       ->  libcusolver
    libcusparse_devel       ->  libcusparse
    libnpp_devel            ->  libnpp
    libnvjpeg_devel         ->  libnvjpeg
    nvidia_driver_devel     ->  nvidia_driver   
    nvidia_fabric_manager   ->  nvidia_driver     # special
    nvidia_libXNVCtrl       ->  nvidia_driver     # special
```

recipe/cross_compile_support.sh
h-vetinari: > The RHS then gives you the list of things I'm looping over.

For convenience (without the sed-ing & the whole devel shenanigans), the jq query based on the upstreams versions.json looks like:
```
"cuda-11-2-11.2.2-1"
"cuda_cudart-11-2-11.2.152-1"
"cuda_cuobjdump-11-2-11.2.152-1"
"cuda_cupti-11-2-11.2.152-1"
"cuda_cuxxfilt-11-2-11.2.152-1"
"cuda_gdb-11-2-11.2.152-1"
"cuda_nvcc-11-2-11.2.152-1"
"cuda_nvdisasm-11-2-11.2.152-1"
"cuda_nvml_dev-11-2-11.2.152-1"
"cuda_nvprof-11-2-11.2.152-1"
"cuda_nvprune-11-2-11.2.152-1"
"cuda_nvrtc-11-2-11.2.152-1"
"cuda_nvtx-11-2-11.2.152-1"
"cuda_samples-11-2-11.2.152-1"
"cuda_sanitizer_api-11-2-11.2.152-1"
"libcublas-11-2-11.4.1.1043-1"
"libcufft-11-2-10.4.1.152-1"
"libcurand-11-2-10.2.3.152-1"
"libcusolver-11-2-11.1.0.152-1"
"libcusparse-11-2-11.4.1.1152-1"
"libnpp-11-2-11.3.2.152-1"
"libnvjpeg-11-2-11.4.0.152-1"
"nsight_compute-11-2-2020.3.1.4-1"
"nsight_systems-11-2-2020.4.3.7-1"
"nvidia_driver-11-2-460.32.03-1"
```

h-vetinari: ... which now makes me realise that 

```
"nsight_compute-11-2-2020.3.1.4-1"
"nsight_systems-11-2-2020.4.3.7-1"
"nvidia_driver-11-2-460.32.03-1"
```

are wrong. Which is annoying, because `nsight-compute-2020.3.1-2020.3.1.4-1.aarch64.rpm` is constructed from a different naming scheme than the libcu* packages (which always have "11-2" in there, even if the version is "11.4.x", c.f. `libcublas-11-2-11.4.1.1043-1.aarch64.rpm`)

recipe/cross_compile_support.sh
```diff
@@ -15,19 +22,102 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+                    "nvidia_fabric_manager:nvidia_driver"
+                    "nvidia_fabric_manager_devel:nvidia_driver"
+                    "nvidia_libXNVCtrl:nvidia_driver"
+                    "nvidia_libXNVCtrl_devel:nvidia_driver"
+                )
+                # add xxx-devel package to manifest with same version (and formatting) as xxx
+                for map in "${DEVELS[@]}"; do
+                    new_key=$(echo $map | cut -d ':' -f1)
+                    from_old=$(echo $map | cut -d ':' -f2)
+                    DEVELQUERY="${DEVELQUERY:+${DEVELQUERY} |}} . += {\"${new_key}\": {\"version\": .${from_old}.version}}"
```
h-vetinari, line 79:
```suggestion
                    DEVELQUERY="${DEVELQUERY:+${DEVELQUERY} \|}} . += {\"${new_key}\": {\"version\": .${from_old}.version}}"
```

recipe/cross_compile_support.sh
```diff
@@ -15,19 +22,102 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
...
+                    "nvidia_fabric_manager:nvidia_driver"
+                    "nvidia_fabric_manager_devel:nvidia_driver"
+                    "nvidia_libXNVCtrl:nvidia_driver"
+                    "nvidia_libXNVCtrl_devel:nvidia_driver"
+                )
+                # add xxx-devel package to manifest with same version (and formatting) as xxx
+                for map in "${DEVELS[@]}"; do
+                    new_key=$(echo $map | cut -d ':' -f1)
+                    from_old=$(echo $map | cut -d ':' -f2)
+                    DEVELQUERY="${DEVELQUERY:+${DEVELQUERY} |} . += {\"${new_key}\": {\"version\": .${from_old}.version}}"
```
h-vetinari, line 79:
```suggestion
                    DEVELQUERY="${DEVELQUERY:+${DEVELQUERY} |} . += {${new_key}: {version: .${from_old}.version}}"
```

recipe/cross_compile_support.sh
kkraus14: I was really hoping we could avoid needing the actual runtime libraries and could get away with just getting the stub libraries but it looks like things aren't packaged in a way to enable that :cry:

Thanks for the explanation. That looks good!

isuruf (commented):
This is way more complicated and downloads more things than needed. The packages I had in my PR were the only ones in `$CUDA_HOME/targets/sbsa-linux` which is what the cross compiler looks at. All the other libraries are irrelevant.

recipe/cross_compile_support.sh
isuruf: Those are not needed.

h-vetinari:
> This is way more complicated and downloads more things than needed.

Fair enough, I was just following the instructions to download packages based on the manifest. I started with downloading everything (+ what Keith told me to add), but if we can filter/shorten the list, all the better

recipe/cross_compile_support.sh
h-vetinari: Can you specify what you mean by "those"? The original list was based on the devel-libs you had in your PR, but weren't listed in the manifest

kkraus14: `nvidia_libXNVCtrl` is GPL licensed (https://github.com/NVIDIA/nvidia-settings/blob/main/COPYING) and not governed under the EULA so I think (I am not a lawyer!) we can actually ship that as a conda package if desired so that can definitely be removed from the list in hindsight. My apologies.

I can't find a license or EULA for the `nvidia_fabric_manager` (should actually have been `nvidia_fabric_manager_devel`, my bad again), so I assumed we weren't allowed to ship it as a conda package. I haven't seen anyone actually use this library before so I'd be perfectly happy to leave it out.

There's other things that would potentially be needed that there aren't stub libraries available for:
- `nvidia-driver-libs` notably contains `libnvoptix` and a bunch of other libraries
- `nvidia-driver-cuda-libs` notably contains `libnvidia-nvvm` and `libnvidia-ptxjitcompiler` and more generally contains the libraries that the symlinks in `nvidia_driver_devel` point to

I have seen a fair amount of software that uses these options where I think it would make sense to add these libraries: `nvidia_driver_devel`, `nvidia_driver_libs`, and `nvidia_driver_cuda_libs`.

Notably with the proposed above, we'll be reliant on stubs for the CUDA driver library `libcuda` and the NVIDIA Management Library `libnvidia-ml`. I'm not sure how build systems like CMake will react to using a mix of actual and stub libraries. Hopefully everything will be okay?

kkraus14:
> This is way more complicated and downloads more things than needed. The packages I had in my PR were the only ones in `$CUDA_HOME/targets/sbsa-linux` which is what the cross compiler looks at. All the other libraries are irrelevant.

I think there's a handful of other libraries that end up in the sysroot (I think?) that there aren't stubs for that we'd want as well that weren't included in your PR. I did my best to break them down in my comment in the thread.

recipe/cross_compile_support.sh
h-vetinari: `nvidia-libXNVCtrl{,-devel}` and `nvidia-fabric-manager{,-devel}` don't exist under:
https://developer.download.nvidia.com/compute/cuda/repos/rhel8/cross-linux-sbsa/
https://developer.download.nvidia.com/compute/cuda/repos/rhel8/ppc64le/

h-vetinari:
> @isuruf: This is way more complicated and downloads more things than needed.

> @kkraus14: There's other things that would potentially be needed [...]

I'll let you and Isuru decide what should be included.

My goal here was just to make it possible to use the versions from the manifest. As feared, this turned into a fair amount of code, due to various small and not so small divergences between the manifest and the actual RPMs. It's conceivable that someone might prefer to just hardcode the list of deps, but now that it works (w/ ability to map in additional rpms, and the obvious possibility to filter out unwanted packages), I tend to think it's probably a good thing going forward.

@isuruf, I managed to get this to run through to the point where you had added
```
mv ./usr/local/cuda-${CUDA_COMPILER_VERSION}/targets/${CUDA_HOST_PLATFORM_ARCH}-linux ${CUDA_HOME}/targets/${CUDA_HOST_PLATFORM_ARCH}-linux
```
at the end of the loop, but this fails with
```
mv: cannot move ‘./usr/local/cuda-11.2/targets/sbsa-linux’ to ‘/usr/local/cuda/targets/sbsa-linux’: Permission denied
```

h-vetinari:
@isuruf @kkraus14 
Can we iterate on the list of things to install?

At the moment the list is as follows:

<details>

```
{
  "cuda": {
    "name": "CUDA SDK",
    "version": "11.2.2"
  },
  "cuda_cudart": {
    "name": "CUDA Runtime (cudart)",
    "version": "11.2.152"
  },
  "cuda_cuobjdump": {
    "name": "cuobjdump",
    "version": "11.2.152"
  },
  "cuda_cupti": {
    "name": "CUPTI",
    "version": "11.2.152"
  },
  "cuda_cuxxfilt": {
    "name": "CUDA cu++ filt",
    "version": "11.2.152"
  },
  "cuda_gdb": {
    "name": "CUDA GDB",
    "version": "11.2.152"
  },
  "cuda_nvcc": {
    "name": "CUDA NVCC",
    "version": "11.2.152"
  },
  "cuda_nvdisasm": {
    "name": "CUDA nvdisasm",
    "version": "11.2.152"
  },
  "cuda_nvml_dev": {
    "name": "CUDA NVML Headers",
    "version": "11.2.152"
  },
  "cuda_nvprof": {
    "name": "CUDA nvprof",
    "version": "11.2.152"
  },
  "cuda_nvprune": {
    "name": "CUDA nvprune",
    "version": "11.2.152"
  },
  "cuda_nvrtc": {
    "name": "CUDA NVRTC",
    "version": "11.2.152"
  },
  "cuda_nvtx": {
    "name": "CUDA NVTX",
    "version": "11.2.152"
  },
  "cuda_samples": {
    "name": "CUDA Samples",
    "version": "11.2.152"
  },
  "cuda_sanitizer_api": {
    "name": "CUDA Compute Sanitizer API",
    "version": "11.2.152"
  },
  "libcublas": {
    "name": "CUDA cuBLAS",
    "version": "11.4.1.1043"
  },
  "libcufft": {
    "name": "CUDA cuFFT",
    "version": "10.4.1.152"
  },
  "libcurand": {
    "name": "CUDA cuRAND",
    "version": "10.2.3.152"
  },
  "libcusolver": {
    "name": "CUDA cuSOLVER",
    "version": "11.1.0.152"
  },
  "libcusparse": {
    "name": "CUDA cuSPARSE",
    "version": "11.4.1.1152"
  },
  "libnpp": {
    "name": "CUDA NPP",
    "version": "11.3.2.152"
  },
  "libnvjpeg": {
    "name": "CUDA nvJPEG",
    "version": "11.4.0.152"
  },
  "nsight_compute": {
    "name": "Nsight Compute",
    "version": "2020.3.1.4"
  },
  "nsight_systems": {
    "name": "Nsight Systems",
    "version": "2020.4.3.7"
  },
  "nvidia_driver": {
    "name": "NVIDIA Linux Driver",
    "version": "460.32.03"
  },
  # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  # from manifest up until this point;
  # entries below added by mapping
  # vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
  "cuda_cudart_devel": {
    "version": "11.2.152"
  },
  "cuda_driver_devel": {
    "version": "11.2.152"
  },
  "cuda_nvrtc_devel": {
    "version": "11.2.152"
  },
  "libcublas_devel": {
    "version": "11.4.1.1043"
  },
  "libcufft_devel": {
    "version": "10.4.1.152"
  },
  "libcurand_devel": {
    "version": "10.2.3.152"
  },
  "libcusolver_devel": {
    "version": "11.1.0.152"
  },
  "libcusparse_devel": {
    "version": "11.4.1.1152"
  },
  "libnpp_devel": {
    "version": "11.3.2.152"
  },
  "libnvjpeg_devel": {
    "version": "11.4.0.152"
  },
  "nvidia_driver_devel": {
    "version": "460.32.03"
  }
}
```

</details>

h-vetinari:
More compact list:
```
cuda:11.2.2
cuda-cudart:11.2.152
cuda-cudart-devel:11.2.152
cuda-cuobjdump:11.2.152
cuda-cupti:11.2.152
cuda-cuxxfilt:11.2.152
cuda-driver-devel:11.2.152
cuda-gdb:11.2.152
cuda-nvcc:11.2.152
cuda-nvdisasm:11.2.152
cuda-nvml-devel:11.2.152
cuda-nvprof:11.2.152
cuda-nvprune:11.2.152
cuda-nvrtc:11.2.152
cuda-nvrtc-devel:11.2.152
cuda-nvtx:11.2.152
cuda-samples:11.2.152
cuda-sanitizer:11.2.152
libcublas:11.4.1.1043
libcublas-devel:11.4.1.1043
libcufft:10.4.1.152
libcufft-devel:10.4.1.152
libcurand:10.2.3.152
libcurand-devel:10.2.3.152
libcusolver:11.1.0.152
libcusolver-devel:11.1.0.152
libcusparse:11.4.1.1152
libcusparse-devel:11.4.1.1152
libnpp:11.3.2.152
libnpp-devel:11.3.2.152
libnvjpeg:11.4.0.152
libnvjpeg-devel:11.4.0.152
nsight-compute:2020.3.1.4
nsight-systems:2020.4.3.7
nvidia-driver:460.32.03
nvidia-driver-devel:460.32.03
```

h-vetinari:
> @isuruf, I managed to get this to run through to the point where you had added
> 
> ```
> mv ./usr/local/cuda-${CUDA_COMPILER_VERSION}/targets/${CUDA_HOST_PLATFORM_ARCH}-linux ${CUDA_HOME}/targets/${CUDA_HOST_PLATFORM_ARCH}-linux
> ```
> 
> at the end of the loop, but this fails with
> 
> ```
> mv: cannot move ‘./usr/local/cuda-11.2/targets/sbsa-linux’ to ‘/usr/local/cuda/targets/sbsa-linux’: Permission denied
> ```

I'm pretty sure now that this happens because I was running this in the recipe-section for debugging purposes, where we don't actually have root rights. I think that if this script is run where it should, that actually should run through. In any case, that it works up until this point shows that the downloads are fine, which means I'm marking this as ready (happy to clean up commit history if desired).

.ci_support/migrations/cuda_112_ppc64le_aarch64.yaml
```diff
@@ -0,0 +1,46 @@
...
+cdt_name:  # [ppc64le or aarch64]
+  - cos7   # [ppc64le or aarch64]
+
+docker_image:                                           # [os.environ.get("BUILD_PLATFORM", "").startswith("linux") and (ppc64le or aarch64)]
+   # case: native compilation (build == target)
+   - quay.io/condaforge/linux-anvil-ppc64le-cuda:11.2   # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
+   - quay.io/condaforge/linux-anvil-aarch64-cuda:11.2   # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
+   # case: cross-compilation (build != target)
+   - quay.io/condaforge/linux-anvil-cuda:11.2           # [ppc64le and os.environ.get("BUILD_PLATFORM") == "linux-64"]
+   - quay.io/condaforge/linux-anvil-cuda:11.2           # [aarch64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
```
h-vetinari, line 46: Not sure if adding this migration is strictly speaking necessary, but since the linux builds have both cuda / non-cuda, I think we should do the same for aarch/ppc?

This reflects the changes in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3624, which we should ideally merge before-hand (then we can skip the `use_local: true` here).

h-vetinari:
> I'm pretty sure now that this happens because I was running this in the recipe-section for debugging purposes, where we don't actually have root rights. I think that if this script is run where it should, that actually should run through.

This turned out to be wrong... So now I'm not sure what we need to tweak (chown?) to get the packages into `$CUDA_HOME`. It's a "simple" permission issue, but I don't know well enough which layer here is allowed to do what.

Also worth noting - not everything gets installed in `./usr/local/cuda-11.2`,  some goes in `./opt/nvidia` or `./usr/share`.

h-vetinari:
@isuruf @kkraus14 rebased after #211 :)

h-vetinari:
@isuruf @kkraus14 
Is there something I can do to move this forward? There's been a couple of points (see comments above) over the last few days where I'd appreciate your insights/help.

h-vetinari:
Gentle ping @kkraus14 @isuruf

kkraus14:
I'm perfectly happy if we want to reduce it to just the list of packages that @isuruf mentioned to start and then consider expanding as desired from there.

isuruf:
Please use the list of package names from my original PR. The others cannot be used for cross compiling and adding complicated logic here to download them (for eg: the special case for nsight, etc) makes no sense for me.

kkraus14:
> Please use the list of package names from my original PR. The others cannot be used for cross compiling and adding complicated logic here to download them (for eg: the special case for nsight, etc) makes no sense for me.

I apologize. I obviously don't know what I'm doing when it comes to cross compilation 😅

h-vetinari:
> Please use the list of package names from my original PR.

In that PR the instruction was to parse the information from the JSON (unclear to me at the time that this wasn't supposed to take all packages from the JSON); just to understand we're talking about the same thing now:
--> the goal is still to take the versions from the JSON, but filter to the packages from your PR <--

?

isuruf:
>  --> the goal is still to take the versions from the JSON, but filter to the packages from your PR

Yes

h-vetinari:
> Please use the list of package names from my original PR.

This now matches the list from #209 (modulo the versions of course, which was the point):
```
+++ echo 'Installing the following packages (<pkg>:<version>)'
Installing the following packages (<pkg>:<version>)
+++ cat rpms_cc.txt
cuda-cudart:11.2.152
cuda-cudart-devel:11.2.152
cuda-cupti:11.2.152
cuda-driver-devel:11.2.152
cuda-nvcc:11.2.152
cuda-nvml-devel:11.2.152
cuda-nvprof:11.2.152
cuda-nvrtc:11.2.152
cuda-nvrtc-devel:11.2.152
cuda-nvtx:11.2.152
libcublas:11.4.1.1043
libcublas-devel:11.4.1.1043
libcufft:10.4.1.152
libcufft-devel:10.4.1.152
libcurand:10.2.3.152
libcurand-devel:10.2.3.152
libcusolver:11.1.0.152
libcusolver-devel:11.1.0.152
libcusparse:11.4.1.1152
libcusparse-devel:11.4.1.1152
libnpp:11.3.2.152
libnpp-devel:11.3.2.152
libnvjpeg:11.4.0.152
libnvjpeg-devel:11.4.0.152
```

Permission errors for 
```
mv ./usr/local/cuda-${CUDA_COMPILER_VERSION}/targets/${CUDA_HOST_PLATFORM_ARCH}-linux ${CUDA_HOME}/targets/${CUDA_HOST_PLATFORM_ARCH}-linux
```
are still there of course...

h-vetinari:
@isuruf
I notice that in #209 the CI was green, but I believe that's because the script wasn't actually run (at least it wasn't executable). I can revert https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210/commits/19330637b189bf19de0008eb0eebc4eca4b09111 of course, but not sure that would solve things in practice.

h-vetinari:
OK, the executable bit was not relevant here after all, but I don't get why this runs into permission errors then...

h-vetinari:
@isuruf Any pointers about the permission errors for moving stuff to `$CUDA_HOME`?

h-vetinari:
Gentle ping @isuruf

h-vetinari:
@isuruf
Should we push this over the finish line? Would be good to re-enable aarch+CUDA builds for arrow.

Or do you think we should wait (several month I'm guessing...?) for https://github.com/conda-forge/staged-recipes/issues/21382 to be finished?

jakirkham:
The compiler pack is already in a staged-recipes PR ( https://github.com/conda-forge/staged-recipes/pull/21350 ) and we are discussing how to support cross-compilation with it. Feedback would be welcome :)

There's no reason it needs to take several months to add the new packages. Though suppose it could take some time. It's more dependent on reviews than anything else. We already have a bunch of recipes ready to go

isuruf:
It adds 11.8 though. We need 11.2. Are there plans to add 11.2?

h-vetinari:
> There's no reason it needs to take several months to add the new packages. Though suppose it could take some time. It's more dependent on reviews than anything else.

My estimate is not meant to downplay how hard people are working on this, but getting 40+ complicated, low-level packages that further interact with each other packaged up is a large undertaking, and I'm doubtful it can be done more quickly than "months", but would be happily surprised if so. In general, very happy to see this effort happening at all! :)

h-vetinari:
Ping @jakirkham @isuruf 
At the december core meeting we had said that this is close, but I don't know how to fix those permission errors. Any ideas/instructions you could share?

h-vetinari:
Seems that https://github.com/conda-forge/docker-images/pull/229 was not enough unfortunately:
```
+++ mv ./usr/local/cuda-11.2/targets/sbsa-linux /usr/local/cuda/targets/sbsa-linux
mv: cannot move ‘./usr/local/cuda-11.2/targets/sbsa-linux’ to ‘/usr/local/cuda/targets/sbsa-linux’: Permission denied
```

(I checked that the images being pulled in are the newest ones from [here](https://quay.io/repository/condaforge/linux-anvil-cuda?tab=tags))

h-vetinari:
OK cool, thanks, this progressed to the next stage now. New boss fight:
```
The reported errors are:
- Encountered problems while solving:
-   - nothing provides requested nvcc_linux-aarch64 11.2.*
```

.ci_support/migrations/cuda_112_ppc64le_aarch64.yaml
h-vetinari: No problem, I'll remove it; though I wonder if we'll still be able to cross-compile cuda without this (which is the point of this PR, no?)

isuruf: How do you propose we solve the part in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210#issuecomment-1412993900?

isuruf:
Also my PR has 3 commits and here there's only 2. Please keep my commit history.

h-vetinari:
> Also my PR has 3 commits and here there's only 2 here. Please keep my commit history.

I squashed together what was a pretty obvious fix-up with its semantic parent.
* 0d2cb012a8dc7f8a849290aed3729575937ddd99
* 06f32ab17dcf9feb532e85b9257cd3c849dbf314

Not least since I've had to rebase this a bunch of different times (_and_ it's a cleaner history), I find this extremely nit-picky. You're free to pick my commits into your PR (squashed, non-squashed, as you wish).

.ci_support/migrations/cuda_112_ppc64le_aarch64.yaml
h-vetinari: > How do you propose we solve the issue in [#210 (comment)](https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210#issuecomment-1412993900)?

I don't know yet, I'm trying to help work through the issues as they appear. But the overarching goal remains to cross-compile CUDA, so if we're falling short of that, we should keep this open IMO.

isuruf:
> Not least since I've had to rebase this a bunch of different times (and it's a cleaner history), I find this extremely nit-picky.

It's not you. I don't like my commits being squashed at all. This has been a struggle trying to get people to not do, but I think I've convinced everybody in conda-forge/core at least to not squash my commits.

h-vetinari:
> This has been a struggle trying to get people to not do, but I think I've convinced everybody in conda-forge/core at least to not squash my commits.

OK, I was not aware of that. Happy to oblige in the future. But since I had to rebase them to keep up with main, the committer will still be me - is this what you want to avoid as well, or just squashing commits together in terms of content?

isuruf:
The committer being you is fine. I just don't want my commit squashed.

h-vetinari:
> The committer being you is fine. I just don't want my commit squashed.

OK, I'll keep this in mind. For context, I often need to trawl through feedstock history, and I find sequences of commits like:
* https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210/commits/07a4edd41aa7e68cc16071e63adccd1abe9c6856
* https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210/commits/19c204f8d5797e5b64d045d2f25840441491f0bd
* https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210/commits/3987006f3940f67585491a2b04a6cc6f58240d25

really annoying (aside from the content-free commit messages), because they tackle one specific issue and _semantically_ should be grouped together. This is not a dig at your commits here, I do the same during development as well, I just have the habit of cleaning up my commit history before merging, so that it's easier to navigate down the line.

But now that I know that you don't like that, I'll make an exception for your commits.

isuruf:
If you want, you can clean up the commit message when there's no message, but keep my commits as they are in the future.

h-vetinari:
> > @isuruf: How do you propose we solve the issue in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210#issuecomment-1412993900?
> 
> @h-vetinari: I don't know yet, I'm trying to help work through the issues as they appear. But the overarching goal remains to cross-compile CUDA, so if we're falling short of that, we should keep this open IMO.

So what are the next steps we need to tackle for cross-compiling CUDA, and where do we track them (if not here)?

isuruf:
We can track them in nvcc-feedstock

recipe/cross_compile_support.sh
```diff
@@ -1,33 +1,123 @@
...
+            # a stable cos8 replacement.
+            EXTRACT_DIR=$(mktemp -d)
+            pushd ${EXTRACT_DIR}
+                if [[ "${HOST_PLATFORM_ARCH}" == "aarch64" ]]; then
+                    CUDA_HOST_PLATFORM_ARCH=sbsa
+                else
+                    CUDA_HOST_PLATFORM_ARCH=${HOST_PLATFORM_ARCH}
+                fi
+                # download manifest for latest 11.2.x patch version
+                curl -L https://developer.download.nvidia.com/compute/cuda/repos/rhel8/${CUDA_HOST_PLATFORM_ARCH}/version_11.2.2.json > manifest.json
```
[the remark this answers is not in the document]
isuruf: You didn't like baking the versions in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/209#discussion_r1010995592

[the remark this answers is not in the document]
kkraus14: I think I was in the mindset that there was an `11.2` manifest that would always give the latest patch version, but I was mistaken. Either way this is fine and we can fix it if there's network hiccups or whatnot in the future.

kkraus14 (approved):

PR #235, opened 2023-04-18
fix typo

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #234, opened 2023-04-18
Install libcuda.so.1 to QEMU sysroot

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

cc @h-vetinari

PR #236, opened 2023-04-18
Cross cuda CI

h-vetinari:
Follow-up to #234 & #235, trying to add CI that exercises `cross_compile_support.sh` - even though we don't exactly need those packages for the infra to run, I think it's better to have them tested on the feedstock of their origin.


varlackc (approved):
LGTM

h-vetinari:
Note: I'm aware that we don't _need_ the CUDA12 and CUDA cross-compilation builds here for our infrastructure to run. But we're not _testing_ those setups here currently, which is what this PR is suggesting to add (especially now that we dropped the CUDA 12 builds).

h-vetinari:
This was essentially done by #261 now. Having cuda 12 jobs is less relevant because https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/main/recipe/cross_compile_support.sh doesn't do much for that version.

PR #241, opened 2023-05-17
Force metadata refresh on upload

jakirkham:
Follow up to discussion ( https://github.com/conda/infrastructure/discussions/649 )

Add an option to force metadata refresh

<hr>

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "3.31.0" %}
+{% set version = "3.31.1" %}
 {% set build = 0 %}
 {% set cuda_compiler_version = cuda_compiler_version or "None" %}
 {% if cuda_compiler_version == "None" %}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -62,7 +62,7 @@ requirements:
     - conda-env
     - click
     - jinja2
-    - anaconda-client >=1.11.0
+    - anaconda-client >=1.11.2
     - conda-build 3.*
     - shyaml
     - jq  # [unix]
--- a/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
+++ b/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
@@ -100,7 +100,7 @@ def built_distribution_already_exists(cli, name, version, fname, owner):
     return exists
 
 
-def upload(token_fn, path, owner, channels, private_upload=False, force_metadata_update=False):
+def upload(token_fn, path, owner, channels, private_upload=False, force_metadata_update=True):
     cmd = ['anaconda', '--quiet', '--show-traceback', '-t', token_fn,
            'upload', path, '--user={}'.format(owner),
            '--channel={}'.format(channels)]
--- a/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
+++ b/recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
@@ -100,12 +100,14 @@ def built_distribution_already_exists(cli, name, version, fname, owner):
     return exists
 
 
-def upload(token_fn, path, owner, channels, private_upload=False):
+def upload(token_fn, path, owner, channels, private_upload=False, force_metadata_update=False):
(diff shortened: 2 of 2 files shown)


jakirkham:
@beckermr, could you please take a look and let us know whether this is in the right direction or what else we might need to do?

jakirkham:
cc @leofang (for awareness)

beckermr (changes requested):
This will only update the metadata on cf-staging IIUIC.

leofang:
Thanks, John. Can a feedstock maintainer flip the switch `force_metadata_update`?

beckermr:
We upload to cf-staging and then copy from staging to conda-forge. We need them to force the refresh in that copy operatoon.

leofang:
> We upload to cf-staging and then copy from staging to conda-forge. We need them to force the refresh in that copy operatoon.

I thought @jessewiles [was saying](https://github.com/conda/infrastructure/discussions/649#discussioncomment-5549747) that whatever metadata would be copied over?

beckermr:
Ah ok. We can try it. My recollection was that this also did not work since for cf-staging we would delete packages regularly and so they should have gotten fresh metadata in any case before the copy IIUIC.

jakirkham:
Right that was my understanding of Jesse's comment as well

Though also was a little confused so [asked for explicit clarification on that point]( https://github.com/conda/infrastructure/discussions/649#discussioncomment-5557083 ), which [Jesse confirmed]( https://github.com/conda/infrastructure/discussions/649#discussioncomment-5560148 )

Ofc the real testing of this is trying it out (and we could of course still find issues), which we could then iterate on with Anaconda 🙂

<hr>

How often do we delete packages? Wonder if deleting a package would be sufficient or if we would need to delete the whole entry it lives under (not sure what the anaconda-client API calls this)

beckermr:
My vote is we set this to true by default since this is really a bug in anaconda.org imho.

jakirkham:
Would be ok with that as well. Originally was just going to set it to `True` in a bunch of places

Do you have thoughts on behavior if the `anaconda-client` is too old? Should we warn and fallback to current behavior? Raise an error? Something else?

Didn't quite understand how we were using the `private` key here. Guessing this is used by others outside of conda-forge as part of feedstocks they manage. Do we need to consider anything for their case or can that be ignored?

beckermr:
Yes some people use our tooling outside of cf.

beckermr:
We should add a requirement on the anaconda-client version in the recipe to ensure it is not too old.

jakirkham:

recipe/meta.yaml
```diff
@@ -62,7 +62,7 @@ requirements:
     - conda-env
     - click
     - jinja2
-    - anaconda-client >=1.11.0
+    - anaconda-client >=1.11.2
```
jakirkham, line 65: xref: https://github.com/Anaconda-Platform/anaconda-client/commit/d0ae59f653845eb40b9a2c2296855e003a382dad

beckermr:
@conda-forge-admin restart ci

leofang:
Nice, thank you guys. Should I kick off a rebuild to test it?

beckermr:
Sure go for it. You'll need to bump the build number.

leofang:
Nope, unfortunately it's still not fixed. Although this is a multi-output recipe, so the behavior could be different? 
截圖 2023-06-13 上午11 11 27

leofang:
it seems it's fixed on cf-staging, so it's just that the copy does not propagate the updated metadata
截圖 2023-06-13 上午11 33 00

jakirkham:
@leofang thanks for trying! 🙏

Could you please share this feedback in the Conda discussion ( https://github.com/conda/infrastructure/discussions/649 )?

beckermr:
OK. So it is like we thought.

beckermr:
Does this call overwrite everything or just metadata? We should not risk nuking existing packages if we can avoid it.

jakirkham:
Let's discuss in that PR for simplicity (where this has [already been answered]( https://github.com/conda-forge/conda-forge-webservices/pull/559#discussion_r1228771188 ))

beckermr:
yep sorry!

jakirkham:
No worries and thank you for taking a look! 🙏

PR #250, opened 2023-07-14
Allow cross-compilation with CUDA 12

h-vetinari:
Trying to use CUDA 12 in https://github.com/conda-forge/arrow-cpp-feedstock/pull/1120, and it fails unsurprisingly because the branches here make it so.

Also picking up a minor comment improvement from #236

### DIFF
--- a/recipe/cross_compile_support.sh
+++ b/recipe/cross_compile_support.sh
@@ -118,10 +118,16 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
         elif [[ "${CUDA_COMPILER_VERSION}" == "11.2" ]]; then
             echo "cross compiling with cuda == 11.2 and cdt != cos7/8 not supported yet"
             exit 1
+        elif [[ "${CUDA_COMPILER_VERSION}" == "12.0" ]] && [[ "${CDT_NAME}" == "cos7" ]]; then
+            # No extra steps necessary for CUDA 12, handled through new packages
+            true
+        elif [[ "${CUDA_COMPILER_VERSION}" == "12.0" ]]; then
+            echo "cross compiling with cuda == 12.0 and cdt != cos7 not supported yet"
+            exit 1
         elif [[ "${CUDA_COMPILER_VERSION}" != "None" ]]; then
             # FIXME: can use anaconda.org/nvidia packages to get the includes and libs
             # for cuda >=11.3.
-            echo "cross compiling with cuda != 11.2 not supported yet"
+            echo "cross compiling with cuda not in (11.2, 12.0) not supported yet"
             exit 1
         fi
     fi
--- a/recipe/cross_compile_support.sh
+++ b/recipe/cross_compile_support.sh
@@ -116,13 +116,13 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
             popd
             rm -rf ${EXTRACT_DIR}
         elif [[ "${CUDA_COMPILER_VERSION}" == "11.2" ]]; then
-	    echo "cross compiling with cuda == 11.2 and cdt != cos7/8 not supported yet"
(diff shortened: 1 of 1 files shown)


recipe/cross_compile_support.sh
```diff
@@ -115,13 +116,19 @@ if [[ "${HOST_PLATFORM}" != "${BUILD_PLATFORM}" ]]; then
             popd
             rm -rf ${EXTRACT_DIR}
         elif [[ "${CUDA_COMPILER_VERSION}" == "11.2" ]]; then
-	    echo "cross compiling with cuda == 11.2 and cdt != cos7/8 not supported yet"
-	    exit 1
+            echo "cross compiling with cuda == 11.2 and cdt != cos7/8 not supported yet"
+            exit 1
+        elif [[ "${CUDA_COMPILER_VERSION}" == "12.0" ]] && [[ "${CDT_NAME}" == "cos7" ]]; then
+            # Need libcuda.so.1 in test phase to test the executables
+            mv ./usr/local/cuda-${CUDA_COMPILER_VERSION}/compat/* ${QEMU_LD_PREFIX}/usr/lib/
```
h-vetinari, line 123: Not yet sure what we need to do here TBH. But the even the CUDA 12 setup doesn't contain `libcuda.so` anywhere, so I think this is still necessary. Downloading the artefacts shouldn't be necessary anymore, but we probably have to add dependencies on the cuda-compiler stack.

h-vetinari:
```
mv: cannot stat ‘./usr/local/cuda-12.0/compat/*’: No such file or directory
```
We'll still have to install CUDA here somehow to get the driver I think

h-vetinari: So basically we just need a branch that doesn't `exit 1` for cuda 12?

jakirkham: Yeah think so. It would be good if this works for any 12

h-vetinari: 12 already works for native compilation, it's just that cross-compilation errors out for that currently

jakirkham: Yeah we actually did some testing offline with cross-compilation and it seemed to work in the test cases we tried

The issues are this line needing to be changed and in some cases NVIDIA packages need GLIBC 2.28 at a minimum, which then become motivation for addressing issue ( https://github.com/conda-forge/conda-forge.github.io/issues/1941 ) in some fashion

That said, I think the Arrow case can skate by without resolving the GLIBC 2.28 (as its usage of CUDA packages is quite minimal). Though it is something we would want to address before migrating CUDA 12 arch packages generally

h-vetinari: Alright, I've set this to do nothing for CUDA 12 now.

h-vetinari:
So this is doing the minimal changes of just not erroring out on CUDA 12 now. In an ideal world, that's all we need. It's possible we still have to iterate a bit, but since no-one is cross-compiling with CUDA 12 yet, this has no risk of breakage, so I'm putting it in. :)

jakirkham:
Think we also need to bump the version number

https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/679cbdce93607a3c76b578165d60b6b67ad8ac75/recipe/meta.yaml#L2

h-vetinari:
Gah, sorry for missing that. 😑

jakirkham:
All good. This feedstock is a bit of a weird one

h-vetinari:
https://github.com/conda-forge/conda-forge-ci-setup-feedstock/commit/17fd53570288308428591cd8c6b04c95a8a70367

Clearly my coffee hasn't kicked in yet. Apologies again.

jakirkham:
Looks good. Thanks! 🙏

Other people have missed the same thing before (self included). So wouldn't worry about it

Curious to see how this goes 🙂

PR #259, opened 2023-08-22
Start using `conda`'s `libmamba` `solver` in CI

jakirkham:
Go ahead and opt-in to using `conda`'s `libmamba` `solver`. This will soon become the default. So it would be good to get a head start using this.

<hr>

Checklist

### DIFF
--- a/recipe/run_conda_forge_build_setup_linux
+++ b/recipe/run_conda_forge_build_setup_linux
@@ -5,6 +5,7 @@ export PYTHONUNBUFFERED=1
 conda config --env --set show_channel_urls true
 conda config --env --set auto_update_conda false
 conda config --env --set add_pip_as_python_dependency false
+conda config --env --set solver libmamba
 # Otherwise packages that don't explicitly pin openssl in their requirements
 # are forced to the newest OpenSSL version, even if their dependencies don't
 # support it.
--- a/recipe/run_conda_forge_build_setup_osx
+++ b/recipe/run_conda_forge_build_setup_osx
@@ -9,6 +9,7 @@ export PYTHONUNBUFFERED=1
 conda config --env --set show_channel_urls true
 conda config --env --set auto_update_conda false
 conda config --env --set add_pip_as_python_dependency false
+conda config --env --set solver libmamba
 # Otherwise packages that don't explicitly pin openssl in their requirements
 # are forced to the newest OpenSSL version, even if their dependencies don't
 # support it.
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -8,6 +8,7 @@ set PYTHONUNBUFFERED=1
 conda.exe config --set show_channel_urls true
 conda.exe config --set auto_update_conda false
 conda.exe config --set add_pip_as_python_dependency false
+conda.exe config --set solver libmamba
 :: Otherwise packages that don't explicitly pin openssl in their requirements
 :: are forced to the newest OpenSSL version, even if their dependencies don't
 :: support it.


jakirkham:
@conda-forge/core please share your thoughts on this 🙂

beckermr:
PR looks great! We do have mamba instead of conda specified in a bunch of spots in smithy and the like. So this PR alone won't fully switch us over. That's fine OFC. I just wanted to mention it for clarity and those following along!

beckermr:
@jaimergp you had a recommendation for a minimum stable version of the libmamba solver. Are we going to be consistent with that by pulling the latest version? Should we hardcode a `>=` constraint somewhere?

jakirkham:
Thanks Matt! 🙏

> We do have mamba instead of conda specified in a bunch of spots in smithy and the like. So this PR alone won't fully switch us over. That's fine OFC. I just wanted to mention it for clarity and those following along!

Definitely true. Wanted to start here to make sure we have this piece handled

It's easy to explore the next step in a conda-smithy PR (and test it in a feedstock) before rolling that change out more generally

> Are we going to be consistent with that by pulling the latest version? Should we hardcode a `>=` constraint somewhere?

Would it be ok to stick with [the version set in the installer]( https://github.com/conda-forge/miniforge/blob/61e7d54c9f3014e0d7a16dc8d30f68378e8b92cd/Miniforge3/construct.yaml#L3 )? Or do we actually need to specify this in more places (like is there risk of it being downgraded)?

beckermr:
Another wrinkle is that we might be upgrading deps and envs on the fly when running jobs even if we have a version in the installer?

jakirkham:
We could run a `libmamba` solve to produce the contents of the installer artifact. Maybe this would ensure those changes are already baked in

jaimergp:
The larger conversation is whether we start using periodically updated lockfiles to provision the installation instead of a blind "update all" that sometimes leaves us in uncertain spots, but yea, a topic for another issue.

jakirkham:
Thanks Jaime! 🙏

So it sounds like there are no action items (for this PR) from your perspective. Or did I missing something?

jaimergp:
Yep, no extra items I can think of. We just need to make sure we use `conda` instead of `mamba` in the right places. To that end I will submit a PR in conda-smithy similar to https://github.com/conda-forge/conda-smithy/pull/1732.

isuruf:
Can you add `conda-libmamba-solver` to `run` requirements?

jaimergp:
I wonder whether we should also provide some configurability like in the strict channel priority setting:

https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/22b4448b3d9feda2f78f94149b68a015366c3a9e/recipe/run_conda_forge_build_setup_linux#L30-L34

, which might require a `conda` namespace in `conda-forge.yml`, similar to what `conda-build` has.

beckermr:
How would that interact with condarc handling?

jaimergp:
Whatever config source is last wins. The precedence is:

- config files, [in this order](https://github.com/conda/conda/blob/e9870ccaaa1ee1c57ff38b075f308e6c051c72a7/conda/base/constants.py#L44-L62): CONDA_ROOT/condarc < user condarc < CONDA_PREFIX/condarc
- environment variables
- command-line flags

What are the concerns for those interactions?

beckermr:
Oh I meant our munging of the condarc. I was hoping we could do all of our changes to it in one spot in our CI system. IDK how many spots we have now.

beckermr:
Also due to the John handles feedstocks, we cannot push to this PR directly. If we want to finish it while he is away, we'll need to move the code to another account that is not an org.

jaimergp:
> IDK how many spots we have now.

Yea, me neither 😂 This will need a deeper audit, but I can think of the following spots in our supply chain right now:

* Docker image (Linux only), which installs...
* ... Miniforge, that ships its own $CONDA_ROOT/condarc with the conda-forge channels.
* AFAIK, conda-smithy does not set any conda settings, but it does allow some configuration (`channel_priority`) for the default values in...
* conda-forge-ci-setup, which invokes `conda config` a few times.

Users could choose between conda-build and mambabuild before too, via `conda-forge.yml`, so it feels natural to allow that configurability here too.

> If we want to finish it while he is away, we'll need to move the code to another account that is not an org.

I'll open a new PR and credit John in the commits.

jaimergp:
Continues on https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/275

jakirkham:
Thanks Jaime! 🙏

Issue #99, opened 2020-05-27
filter outputs for upload/validation by name


beckermr:
It appears that when conda build runs downstream tests it leaves the downstream packages in the build artifacts directory. This causes our code to attempt to validate and upload them, which leads to jobs that fail when they should otherwise pass. We can compute the names of the outputs from the meta.yaml (but not their versions) and filter things we find by globing. 

Cc @isuruf this is why your binutils pr appears to be failing. 

Issue:

<br/>
Environment (<code>conda list</code>):
<details>

```
$ conda list

```
</details>

<br/>
Details about  <code>conda</code> and system ( <code>conda info</code> ):
<details>

```
$ conda info

```
</details>

PR #261, opened 2023-08-23
Add CUDA 11.8 cross-compilation support

jakirkham:
This adds additional logic to the CUDA 11 cross-compilation case here to support CUDA 11.8.

Needed as part of the CUDA 11.8 migration: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4834

<hr>

Checklist

### DIFF
--- a/.ci_support/migrations/cuda118.yaml
+++ b/.ci_support/migrations/cuda118.yaml
@@ -1,0 +1,91 @@
+migrator_ts: 1692828152
+__migrator:
+  kind:
+    version
+  migration_number:
+    1
+  build_number:
+    1
+  paused: false
+  override_cbc_keys:
+    - cuda_compiler_stub
+  operation: key_add
+  check_solvable: false
+  primary_key: cuda_compiler_version
+  ordering:
+    cxx_compiler_version:
+      - 9
+      - 8
+      - 7
+    c_compiler_version:
+      - 9
+      - 8
+      - 7
+    fortran_compiler_version:
+      - 9
+      - 8
+      - 7
+    docker_image:
+      - quay.io/condaforge/linux-anvil-comp7              # [os.environ.get("BUILD_PLATFORM") == "linux-64"]
+      - quay.io/condaforge/linux-anvil-aarch64            # [os.environ.get("BUILD_PLATFORM") == "linux-aarch64"]
+      - quay.io/condaforge/linux-anvil-ppc64le            # [os.environ.get("BUILD_PLATFORM") == "linux-ppc64le"]
+      - quay.io/condaforge/linux-anvil-armv7l             # [os.environ.get("BUILD_PLATFORM") == "linux-armv7l"]
+      - quay.io/condaforge/linux-anvil-cuda:9.2           # [linux64 and os.environ.get("BUILD_PLATFORM") == "linux-64"]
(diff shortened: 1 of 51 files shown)


h-vetinari (commented):
It would be nice if this actually turns out this painless (given the struggle to get this right for 11.2)!

I would still suggest to actually let this run through execution here before merging. At the time of #210, I had locally pulled in a (cross-enabled) 11.2 migrator, so that the CI would generate the respective jobs here and actually execute the branches in question.

In your case, this would be a cross-enabled version of https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4834 (plus `local: True`).

I don't necessarily expect that nvidia's artefact naming has changed a lot between 11.2 & 11.8, but IMO better to double-check. As one concrete issue, the `extra="11-2-"` further down will need adaptation as well.

jakirkham:
Thanks Axel! 🙏

Appreciate the helpful tips. Have tried to push some changes to address those

Please let me know how that looks

.ci_support/migrations/cuda118.yaml
```diff
@@ -2,6 +2,7 @@ migrator_ts: 1692828152
 __migrator:
   kind:
     version
+  use_local: true # TODO: Remove this line after completing testing
```
h-vetinari, line 5: Not sure that we want to have the cross-compiled jobs for merging. We don't _need_ them from the POV of the CI setup itself, just for testing the branches here in CI. I suggested adding them in https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/236, but never got a response.

jakirkham: Yep no strong feelings here. Happy to keep or drop as preferred

jakirkham:
CI is failing as the `nvcc` wrapper script package has not been built for CUDA 11.8 cross-compilation

Addressing with PR: https://github.com/conda-forge/nvcc-feedstock/pull/99

h-vetinari:
Tried to merge main here after #99, but it seem you've unchecked the box that says "allow edits by maintainers"

jakirkham:
Thanks Axel! 🙏

Unfortunately since this is an org fork (as opposed to a user fork) that check box isn't even available (I couldn't check it if I wanted to)

Have clicked the update branch button. So we should be good to go

Also FWIW closing and reopening PRs has the same effect (it deletes the merge ref for the PR and recreates it using the latest upstream branch), which will also update CI as a result

h-vetinari:
Ah, I had forgotten that this depends on https://github.com/conda-forge/nvcc-feedstock/pull/99. 🙃

jakirkham:
Yeah no worries. This wound up being a bit of a yak shave (though that was mostly expected)

h-vetinari:
I did this in #267 and it's passing! 🥳 

If we do not want to keep the cross-compilation jobs (which would be purely for testing purposes), then you could rebase/revert out the migrator and merge this from my POV. 👍

jakirkham:
Thanks Axel! 🙏

Went ahead and cherry-picked the last 2 commits. Hope that is ok

jakirkham:
Forgot to bump version 🤦‍♂️

Fixed with commit: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/commit/22b4448b3d9feda2f78f94149b68a015366c3a9e

PR #292, opened 2023-11-06
Support CUDA 12 + Windows

leofang:
The logic here is to treat CUDA 12+ differently from previous CUDA versions. Specifically, for CUDA 12+ we
- do not download/extract any CUDA installer 
- do not need to set `CUDA_PATH` or update `PATH`
- but still need to set `CONDA_OVERRIDE_CUDA`

Tested in https://github.com/conda-forge/cupy-feedstock/pull/228.

cc: @conda-forge/cuda 

Checklist

### DIFF
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -91,15 +91,23 @@ for %%i in ("%~dp0.") do set "SCRIPT_DIR=%%~fi"
 <cuda.version set /p CUDA_VERSION=
 del cuda.version
 if not "%CUDA_VERSION%" == "None" (
-    call "%SCRIPT_DIR%\install_cuda.bat" %CUDA_VERSION%
-    if errorlevel 1 (
-        echo Could not install CUDA
-        exit 1
+    if "%CUDA_VERSION:~0,2%" == "12" (
+        :: Don't call install_cuda, as we'll get CUDA packages from CF
+        set "CUDA_PATH="
+        set "CONDA_OVERRIDE_CUDA=%CUDA_VERSION%"
+        :: Export CONDA_OVERRIDE_CUDA to allow __cuda to be detected on CI systems without GPUs
+        echo set "CONDA_OVERRIDE_CUDA=%CONDA_OVERRIDE_CUDA%" >> "%CONDA_PREFIX%\etc\conda\activate.d\conda-forge-ci-setup-activate.bat"
+    ) else (
+        call "%SCRIPT_DIR%\install_cuda.bat" %CUDA_VERSION%
+        if errorlevel 1 (
+            echo Could not install CUDA
+            exit 1
+        )
+        :: We succeeded! Export paths
+        set "CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v%CUDA_VERSION%"
+        set "PATH=%PATH%;%CUDA_PATH%\bin"
+        set "CONDA_OVERRIDE_CUDA=%CUDA_VERSION%"
     )
-    :: We succeeded! Export paths
-    set "CUDA_PATH=C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v%CUDA_VERSION%"
-    set "PATH=%PATH%;%CUDA_PATH%\bin"
-    set "CONDA_OVERRIDE_CUDA=%CUDA_VERSION%"
 )
 :: /CUDA
 


recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -91,15 +91,23 @@ for %%i in ("%~dp0.") do set "SCRIPT_DIR=%%~fi"
...
-    call "%SCRIPT_DIR%\install_cuda.bat" %CUDA_VERSION%
-    if errorlevel 1 (
-        echo Could not install CUDA
-        exit 1
+    if "%CUDA_VERSION:~0,2%" == "12" (
+        :: Don't call install_cuda, as we'll get CUDA packages from CF
+        set "CUDA_PATH="
+        set "CONDA_OVERRIDE_CUDA=%CUDA_VERSION%"
+        :: Export CONDA_OVERRIDE_CUDA to allow __cuda to be detected on CI systems without GPUs
+        echo set "CONDA_OVERRIDE_CUDA=%CONDA_OVERRIDE_CUDA%" >> "%CONDA_PREFIX%\etc\conda\activate.d\conda-forge-ci-setup-activate.bat"
```
xhochy, line 99: Why do we need this, too, in the 12.x case? It also looks like it doesn't really do anything as it sets the variable to itself.

xhochy (changes requested):

recipe/run_conda_forge_build_setup_win.bat
leofang: I thought the comment is self-explanatory? Also, this treatment is copied from existing code (see line 133-134 below), I didn't reinvent any wheel 🙂

xhochy: > I thought the comment is self-explanatory?

Not fully. Reading the source again, it makes sense. "Bake-in" instead of "Export" would have made it clear to me but probably not to everyone else. Happy to keep it as-is.

xhochy (approved):

recipe/run_conda_forge_build_setup_win.bat
jakirkham: Basically with how CUDA 11 is handled (both on Linux and Windows), most CUDA packages come from the system. There is a bunch of logic here to handle those CUDA 11 cases

With CUDA 12, we now have Conda packages and so don't need to use that logic. Eventually we could drop the CUDA 11 logic from here when we drop CUDA 11. In the interim, we just need to make sure we "do nothing" on CUDA 12

recipe/run_conda_forge_build_setup_win.bat
```diff
@@ -91,15 +91,23 @@ for %%i in ("%~dp0.") do set "SCRIPT_DIR=%%~fi"
 <cuda.version set /p CUDA_VERSION=
 del cuda.version
 if not "%CUDA_VERSION%" == "None" (
-    call "%SCRIPT_DIR%\install_cuda.bat" %CUDA_VERSION%
-    if errorlevel 1 (
-        echo Could not install CUDA
-        exit 1
+    if "%CUDA_VERSION:~0,2%" == "12" (
```
leofang, line 94: btw I tried to make this line more robust
```suggestion
    for /f "delims=." %%f in ("%CUDA_VERSION%") do (set /a "CUDA_MAJOR_VERSION=%%f")
    if %CUDA_MAJOR_VERSION% geq 12 (
```
but for some reason the CI just hated the change (it worked fine locally), so I didn't use it here... Like John said, in the future we'll drop most of the complex logic here, so I hope this is OK.

jakirkham:
Think we might need to update these lines as well

https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/fe41924428086877a6936f46d5901bd921202f98/recipe/run_conda_forge_build_setup_win.bat#L130-L135

leofang:
Yes, this block is only reachable by CUDA 11 and below (where `CUDA_PATH` is set). For CUDA 12+, this block would be skipped. It's admitted an ugly hack, but my lack of cmd knowledge (and failures to make CI happy even with correct code, as noted above) is preventing me from doing a better job here.

jakirkham:
Ah gotcha. May just be my fault for looking on my phone

jakirkham:
@conda-forge/core , any other thoughts on this change?

jakirkham (approved):

jakirkham:
Thanks all! 🙏

Let's give this a try and follow up on anything else after

Note: Missed we didn't have the version number bump. So did that in commit ( https://github.com/conda-forge/conda-forge-ci-setup-feedstock/commit/902a2beabaff6fd0b912f490744f6f016131c12b )

leofang:
Thanks for catching the version bump, John. Sent a backport PR (#293).

PR #316, opened 2024-04-17
feat: add rattler build support for upload

nichmor:
Checklist
* [] Bumped the build number (if the version is unchanged)

To add rattler-build in [conda-smithy PR](https://github.com/conda-forge/conda-smithy/pull/1876) we need to also add support for it in upload phase. This PR address it.

https://github.com/conda-forge/rattler-build-conda-compat-feedstock/runs/23921634356 < --- upload with new recipe.yaml tested in CI run here


.scripts/build_steps.sh
```diff
@@ -34,9 +34,9 @@ CONDARC
 export CONDA_LIBMAMBA_SOLVER_NO_CHANNELS_FROM_INSTALLED=1
 
 mamba install --update-specs --yes --quiet --channel conda-forge --strict-channel-priority \
-    pip mamba conda-build conda-forge-ci-setup=4
+    pip mamba conda-build conda-forge-ci-setup=4 rattler-build-conda-compat
 mamba update --update-specs --yes --quiet --channel conda-forge --strict-channel-priority \
-    pip mamba conda-build conda-forge-ci-setup=4
+    pip mamba conda-build conda-forge-ci-setup=4 rattler-build-conda-compat
```
nichmor, line 39: until new version is published we need to specify  rattler-build-conda-compat here

xhochy:
cc @0xbe7a

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -31,13 +33,24 @@ def get_built_distribution_names_and_subdirs(recipe_dir, variant):
                 'clobber_sections_file': clobber_file
             }
             break
+    
+    # if feedstock don't have new recipe.yaml
+    # use default conda_build.api.render for meta.yaml
+    if not rattler_has_recipe(recipe_dir):
```
isuruf, line 39: You should use `conda-forge.yml`'s value for configuring build-tool here.

nichmor: Initially I thought about extracting the value from `conda-forge.yaml` but the `retry_upload_or_check` don't pass any feedstocks paths so I have only recipe_dir.
I can try and go up one level from recipe_dir and assume that it will be the feedstock itself, but I'm not sure if it's the right way of doing it.

Another idea is to have additional option for `upload_or_check_non_existence` ( something like --rattler-build ) that will be passed from conda-smithy.

wdyt?

isuruf: I think we should pass in a feedstock_dir to these functions.

nichmor:
Hey! @isuruf 

I've applied your suggestion and now I'm passing & using feedstock_dir to verify presence of recipe.yaml.

Let me know wdyt

isuruf:
Thanks

recipe/conda_forge_ci_setup/upload_or_check_non_existence.py
```diff
@@ -195,8 +211,17 @@ def upload_or_check(
 
     cli = get_server_api(token=token)
 
+    build_tool = CONDA_BUILD
+
+    if feedstock_root and os.path.exists(os.path.join(feedstock_root, "conda-forge.yml")):
+        with open(os.path.join(feedstock_root, "conda-forge.yml")) as f:
```
isuruf, line 217:
```
    if feedstock_root and os.path.exists(forge_yaml:=os.path.join(feedstock_root, "conda-forge.yml")):
        with open(forge_yaml) as f:
```

PR #338, opened 2024-09-09
feat: error if output fails validation according to global setting

beckermr:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #340, opened 2024-09-11
fix: skip output validation if not on

beckermr:
Checklist

This PR fixes the case where the package is used outside conda-forge or conda_forge_output_validation is turned off.

xref: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/338

### DIFF
--- a/recipe/conda_forge_ci_setup/feedstock_outputs.py
+++ b/recipe/conda_forge_ci_setup/feedstock_outputs.py
@@ -11,7 +11,12 @@ from conda_forge_metadata.feedstock_outputs import (
     feedstock_outputs_config,
 )
 
-from .utils import built_distributions, compute_sha256sum, split_pkg
+from .utils import (
+    built_distributions,
+    compute_sha256sum,
+    split_pkg,
+    is_conda_forge_output_validation_on,
+)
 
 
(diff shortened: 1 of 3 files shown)


beckermr:
@JohanMabille This will fix the bug where we run validation even if we are not on conda-forge. However, given the micromamba build workflow, you will likely need to edit the conda-forge.yml to turn off validation on-the-fly. Also, I noticed that some of the jobs there passed because they did not find any distributions to validate, despite this bug. I wonder if that is another bug.

JohanMabille:
Indeed, only linux-64 finds a distribution to validate. We have almost the same behavior in the [feedstock itself](https://github.com/conda-forge/micromamba-feedstock), see the [linux_ppc64](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1018544&view=logs&jobId=ced5d8de-8227-5f3f-33a1-45cf1592c45a&j=8c139a40-380b-5518-bf6b-edcb18cbf319&t=e71f8d0d-13be-528e-b5e7-bfc20e0db699) build for instance.

JohanMabille:
Hi again, sorry for the question but I think I didn't understand it 100%: if this fixes the validationn when building the feedstock outside of conda-forge (by skipping it IIUC the patch), why must I turn off the validation in the micromamba ci?

PR #344, opened 2024-09-18
fix: ensure validation is run on correct files

### LINKED ISSUE #341: some distributions not found
The validation script appears to not be finding some of the distributions the CI pipelines build: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1018544&view=logs&jobId=ced5d8de-8227-5f3f-33a1-45cf1592c45a&j=8c139a40-380b-5518-bf6b-edcb18cbf319&t=e71f8d0d-13be-528e-b5e7-bfc20e0db699

I think this may be related to cross-compilation and using the proper subdir.

### LINKED ISSUE #342: Output validation fails when there are downstream tests
When there are downstreams tests, conda-build downloads the outputs to the same folder that the built packages are.

For example see https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_apis/build/builds/1028622/logs/45
This fails because cython and setuptools are in https://github.com/conda-forge/python-feedstock/blob/main/recipe/meta.yaml#L184-L187

beckermr:
This PR ensures we run output validation only on the packages we built.

closes #342 
closes #341

### DIFF
--- a/recipe/conda_forge_ci_setup/feedstock_outputs.py
+++ b/recipe/conda_forge_ci_setup/feedstock_outputs.py
@@ -14,6 +14,8 @@ from conda_forge_metadata.feedstock_outputs import (
 from .utils import (
     built_distributions,
     compute_sha256sum,
+    determine_build_tool,
+    get_built_distribution_names_and_subdirs,
     split_pkg,
     is_conda_forge_output_validation_on,
 )
@@ -140,7 +142,21 @@ def main(feedstock_name):
     """Validate the feedstock outputs."""
 
     if is_conda_forge_output_validation_on():
-        distributions = [os.path.relpath(p, conda_build.config.croot) for p in built_distributions()]
+        feedstock_root = os.environ.get(
+            "FEEDSTOCK_ROOT",
+            os.getcwd(),
+        )
+        recipe_root = os.path.join(feedstock_root, "recipe")
+        build_tool = determine_build_tool(feedstock_root)
+        allowed_dist_names, allowed_subdirs = get_built_distribution_names_and_subdirs(
+            recipe_root, None, build_tool=build_tool
+        )
+        distributions = [os.path.relpath(p, conda_build.config.croot) for p in built_distributions(subdirs=allowed_subdirs)]
+        distributions = [
+            dist
+            for dist in distributions
+            if any(dist.startswith(allowed + "-") for allowed in allowed_dist_names)
+        ]
         results = is_valid_feedstock_output(feedstock_name, distributions)
 
         print("validation results:\n%s" % json.dumps(results, indent=2), flush=True)
--- a/recipe/conda_forge_ci_setup/feedstock_outputs.py
+++ b/recipe/conda_forge_ci_setup/feedstock_outputs.py
@@ -148,7 +148,7 @@ def is_valid_feedstock_output(project, outputs):
     '-m',
     multiple=True,
     type=click.Path(exists=False, file_okay=True, dir_okay=False),
-    default=None,
+    default=(),
     help="path to conda_build_config.yaml defining your base matrix",
 )
 def main(feedstock_name, recipe_dir, variant):
--- a/recipe/conda_forge_ci_setup/feedstock_outputs.py
+++ b/recipe/conda_forge_ci_setup/feedstock_outputs.py
@@ -14,7 +14,6 @@ from conda_forge_metadata.feedstock_outputs import (
 from .utils import (
     built_distributions,
     compute_sha256sum,
-    determine_build_tool,
     get_built_distribution_names_and_subdirs,
     split_pkg,
     is_conda_forge_output_validation_on,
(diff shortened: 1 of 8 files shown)


beckermr:
cc @isuruf

recipe/conda_forge_ci_setup/feedstock_outputs.py
```diff
@@ -140,7 +141,13 @@ def main(feedstock_name):
     """Validate the feedstock outputs."""
 
     if is_conda_forge_output_validation_on():
-        distributions = [os.path.relpath(p, conda_build.config.croot) for p in built_distributions()]
+        allowed_dist_names, allowed_subdirs = get_built_distribution_names_and_subdirs()
```
isuruf, line 144: This is an expensive operation. Can we cache this on disk using some python library.

recipe/conda_forge_ci_setup/utils.py
```diff
@@ -9,9 +10,68 @@
...
+    if variant is None:
+        if "CONFIG_FILE" in os.environ:
+            variant = [os.environ.get("CONFIG_FILE")]
+        else:
+            variant = [
+                os.path.join(
+                    os.environ.get("CI_SUPPORT", os.path.join(feedstock_root, ".ci_support")),
+                    os.environ.get("CONFIG") + ".yaml"
+                )
+            ]
```
isuruf, line 37: This is fine to have, but we should deprecate these.

beckermr: Yeah it'd be nice to cleanup the set of supplied env vars. It differs by platform and possibly provider in smithy. 

Which ones did you want to deprecate?

recipe/conda_forge_ci_setup/feedstock_outputs.py
beckermr: Yes let me check into that.

recipe/conda_forge_ci_setup/utils.py
isuruf: `recipe_dir` being None and `variant` being None. They should be passed through to this function.

recipe/conda_forge_ci_setup/feedstock_outputs.py
isuruf: example: https://stackoverflow.com/a/23526717/4768820

beckermr: yes this one too: https://pypi.org/project/cache-to-disk/

recipe/conda_forge_ci_setup/utils.py
isuruf: Yes, I understand why there needs to be default values for now. We should fix smithy, so that it is not needed in the future.

recipe/conda_forge_ci_setup/inspect_artifacts.py
```diff
@@ -1,15 +1,29 @@
...
+from .utils import (
+    built_distributions,
+    compute_sha256sum,
+    get_built_distribution_names_and_subdirs,
+    human_readable_bytes,
+)
 
 
 @click.command()
 def main():
```
isuruf, line 17: for eg: `variant` and `recipe_dir` should be added as `click` options and passed through

beckermr: xref: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/345

beckermr:
ok @isuruf I've added caching and I added the options to the CI setup calls. They are not required right now. Once we adjust smithy, we can turn them to be required if we like.

jaimergp:
I think this broke `inspect_artifacts` in `staged-recipes`. Not sure why we need to render the recipes there (again). [See here for Linux logs (fails but error is ignored)](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1031239&view=logs&jobId=6f142865-96c3-535c-b7ea-873d86b887bd&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823):

```
Inspecting artifacts
+ command -v inspect_artifacts
+ inspect_artifacts
Error: non-existent: /home/conda/staged-recipes/recipe
+ echo 'inspect_artifacts needs conda-forge-ci-setup >=4.6.0'
inspect_artifacts needs conda-forge-ci-setup >=4.6.0
```

The `Error: non-existent` string comes from `conda_build.render`. I guess in staged-recipes we can't pass `/recipe` blindly, since the recipe dir depends on the PR contents.

beckermr:
Ah I see. The goal in this pr was to not print inspection for every artifact but only the ones that were new. When there are downstream tests, we inspect the downstreams too which is odd and confusing.

jaimergp:
Got it. I think we raaaarely use downstreams in staged-recipes, right? Should we special case it? Or adjust this so it accepts several directories and pass `recipes/*` from staged-recipes?

PR #346, opened 2024-09-19
fix: allow all distributions to be inspected

beckermr:
This PR fixes a bug where the inspect command broke on staged recipes. I added a flag we can use there to inspect everything.

cc @jaimergp


xhochy (approved):

jaimergp (approved):
Thanks!

PR #350, opened 2024-09-26
fix: allow for feedstock_root to not be set

beckermr:
This PR fixes the rerendering bug where FEEDSTOCK_ROOT is not set and smithy chokes.

PR #351, opened 2024-09-26
fix: attempt to force latest build number for compilers

beckermr:
Checklist


beckermr:
@conda-forge-admin rerender

recipe/meta.yaml
```diff
@@ -55,7 +55,7 @@ requirements:
   build:
      # ensure that all variants of cuda_compiler_version are tested
     - {{ compiler('cuda') }}              # [cuda_compiler_version != "None"]
-    - {{ compiler('c') }}                 # [cuda_compiler_version != "None"]
+    - {{ c_compiler }}_{{ target_platform }} {{ c_compiler_version }} *_4             # [cuda_compiler_version != "None"]
```
jakirkham, line 58: Guessing we would want something like this (as Windows wouldn't need this pin)

```suggestion
    - {{ c_compiler }}_{{ target_platform }} {{ c_compiler_version }} *_4             # [linux and cuda_compiler_version != "None"]
    - {{ compiler('c') }}                                                                # [(not linux) and cuda_compiler_version != "None"]
```

beckermr: Yes but this is just for testing. If this works, we'll want to use a repodata patch.

beckermr:
ouch another rerender error. I suspect something is still broken in how the latest conda-build versions inject variables into the jinja2 templates.

beckermr:
@conda-forge-admin rerender

recipe/meta.yaml
```diff
@@ -21,9 +21,9 @@ build:
   skip: True
   {% endif %}
   script:
-    - cp {{ os.environ["FEEDSTOCK_ROOT"] }}/LICENSE.txt ${RECIPE_DIR}/LICENSE.txt            # [linux]
-    - cp ${RECIPE_DIR}/../LICENSE.txt ${RECIPE_DIR}/LICENSE.txt                              # [osx]
-    - COPY "%RECIPE_DIR%\\..\\LICENSE.txt" "%RECIPE_DIR%\\LICENSE.txt"                       # [win]
+    - cp {{ os.environ.get("FEEDSTOCK_ROOT", "failme") }}/LICENSE.txt ${RECIPE_DIR}/LICENSE.txt  # [linux]
```
jakirkham, line 25: We can just guard the usage if needed

```suggestion
    {% if "FEEDSTOCK_ROOT" in os.environ %}
    - cp {{ os.environ["FEEDSTOCK_ROOT"] }}/LICENSE.txt ${RECIPE_DIR}/LICENSE.txt  # [linux]
    {% endif %}
```

beckermr: Let's move this comment to PR #350

beckermr:
@conda-forge-admin rerender

recipe/meta.yaml
jakirkham: Only responded here as we were having re-rendering issues: https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/351#issuecomment-2375527805

beckermr: yeah that is a new one that likely has the same underlying cause but is for a different variable...

beckermr:
@conda-forge-admin rerender

jakirkham:
Looks promising 🙂

jakirkham:
Continuing discussion in issue: https://github.com/conda-forge/binutils-feedstock/issues/78

beckermr:
Oh fascinating. These builds with the c compiler included appear to pass. OTOH these builds failed: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1037636&view=logs&j=09c6ee00-92ab-5751-f80f-c049ce8ce0ed&t=7a4216ee-c61d-5132-71cd-0d47b3b4b519

jakirkham:
Repeating this comment ( https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/349#issuecomment-2375439341 ) to keep everything in this thread

We noticed the `conda-forge-ci-setup` cross-compiled architectures are failing (for [example]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1037636&view=logs&j=09c6ee00-92ab-5751-f80f-c049ce8ce0ed&t=7a4216ee-c61d-5132-71cd-0d47b3b4b519&l=866 ))

```python
__main__.py: error: argument --sysroot: expected one argument
```

This led us to the `cross-python` package, which has this logic in [its activation script]( https://github.com/conda-forge/cross-python-feedstock/blob/d94169b29f29de1b24dc39576ac675dac8a6d048/recipe/activate-cross-python.sh#L32-L38 ):

```bash
    $BUILD_PREFIX/bin/python -m crossenv $PREFIX/bin/python \
        --sysroot $CONDA_BUILD_SYSROOT \
        --without-pip $BUILD_PREFIX/venv \
        --sysconfigdata-file "$sysconfigdata_fn" \
	--machine ${machine} \
        --cc ${CC:-@CC@} \
        --cxx ${CXX:-@CXX@}
```

So for some reason `$CONDA_BUILD_SYSROOT` is now undefined

Certainly if we add the compilers (as done here), [`CONDA_BUILD_SYSROOT` will be set by the compilers]( https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/d415477224e3e7bb411766c0e488964e481ab7c8/recipe/activate-gcc.sh#L187 )

```bash
  "CONDA_BUILD_SYSROOT,${CONDA_PREFIX}@LIBRARY_PREFIX@/@CHOST@/sysroot" \
```

However that means all users of `cross-python` now need to add the compilers as a build dependency (if they don't already). Is this an expected/intended change?

PR #360, opened 2024-10-03
[rattler-build] Correctly propagate target platform to `host_{arch|platform}`

hadim:
The target platform was wrongly propagated in the conda build Config object in `arch` and `platform` instead of `host_arch` and `host_platform`. 

Sister PR: https://github.com/prefix-dev/rattler-build-conda-compat/pull/59

I encountered the issue when converting a feedstock to rattler at https://github.com/conda-forge/xformers-feedstock/pull/33


beckermr:
Thanks for the PR! Do we need to wait for the other pr to merge first, or can we merge this one right away?

hadim:
They are both required to fix the cross compilation issue but nothing should break if we merge and release now. We can also wait if you prefer.

wolfv:
rattler-build-conda-compat with the fix is also released.

PR #374, opened 2025-01-07
show qemu version when running cross_compile_support.sh

h-vetinari:
I'm running into a possible regression in QEMU (c.f. https://github.com/conda-forge/ctng-compilers-feedstock/pull/147), and I cannot go back to the last passing CI run to see which version was used. So ensure that the version gets logged.


beckermr (approved):

h-vetinari:
Thanks for the quick review!

PR #183, opened 2022-02-11
unset CPU_COUNT for github actions as well

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

cc @wolfv

PR #381, opened 2025-03-14
Set `$CPU_COUNT` only for some Linux CI providers; remove everywhere else

jaimergp:
Checklist
* [X] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [X] Reset the build number to `0` (if the version changed)
* [X] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

We had this `$CPU_COUNT` logic for some older CI providers, like Appveyor (Windows) or Travis (for macOS). As far as I know, we only need this for some virtualized Linux providers where `nproc` (or equivalent) report the number of CPUs in host, not in guest. In those cases, we manually override `CPU_COUNT` with the known number as per the CI provider documentation.

In the absence of CPU_COUNT, conda-build will estimate it automatically via `multiprocessing.cpu_count()` (`sysctl -n hw.logicalcpu` on macOS). We should let that happen unless we do know this method might misreport the number of CPUs. This value will be injected to the build scripts as `CPU_COUNT`.

Of course, feedstocks can always redefine `CPU_COUNT` themselves if the calculated number of CPUs is too high and causes resource starvation, OOM, etc.

### DIFF
--- a/.scripts/run_docker_build.sh
+++ b/.scripts/run_docker_build.sh
@@ -12,7 +12,7 @@ source .scripts/logging_utils.sh
 set -xeo pipefail
 
 THISDIR="$( cd "$( dirname "$0" )" >/dev/null && pwd )"
-PROVIDER_DIR="$(basename $THISDIR)"
+PROVIDER_DIR="$(basename "$THISDIR")"
 
 FEEDSTOCK_ROOT="$( cd "$( dirname "$0" )/.." >/dev/null && pwd )"
 RECIPE_ROOT="${FEEDSTOCK_ROOT}/recipe"
--- a/build-locally.py
+++ b/build-locally.py
@@ -10,6 +10,7 @@ import glob
 import os
 import platform
 import subprocess
+import sys
 from argparse import ArgumentParser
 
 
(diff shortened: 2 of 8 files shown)


jaimergp:
Hm, I had assumed that we were not setting CPU_COUNT for GHA for the self-hosted cases, but maybe the reasons were others? The blame leads to https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/183, where there's no comment on _why_ this was needed. It looks like setting it broke something somewhere?

@isuruf, do you remember the reason? Do we need to unset it in GHA? If so, should this also apply to other OSs? The same logic was not present in macOS and Windows.

isuruf:
Things like `nproc` are not reliable on some CI providers as only a few processors are available and using a higher number leads to the CI job running out of some resource and getting killed. I removed it for github actions to use the conda_buld default for the self-hosted runners.

beckermr (approved):

isuruf (changes requested):
I'd rather we don't use `nproc` as that's equivalent to unsetting,

jaimergp:
So then I was misunderstanding the intent of that `CPU_COUNT` definition. Let me check: are we hardcoding an overridable `CPU_COUNT` value because in some platforms (e.g. virtualization layers) `nproc` might misreport the number of available CPUs and thus starve the runner via OOMs and similar problems?

An absent `CPU_COUNT` in conda-build is dealt with the value of `multiprocessing.cpu_count()` on Linux/Windows and `sysctl -n hw.logicalcpu` on macOS.

So yea, I get it now, but then we need to special case which providers need the override, and not the other way around? I'll commit something with my current understanding for now.

recipe/run_conda_forge_build_setup_osx
```diff
@@ -1,9 +1,5 @@
 #!/bin/bash
```
beckermr, line 2: I thought we wanted to use 3 for OSX still?

jaimergp:
For some reason CPU_COUNT is not set on macOS by conda-build (but it does show up in Windows correctly) 🤷

beckermr:
@jaimergp can we update the PR title and description to reflect what we are doing now after feedback? I am confused on this one.

jaimergp:
@beckermr, updated title and description.

jaimergp:
@isuruf, can you take another look? I think the current changeset alleviates your concerns.

PR #407, opened 2025-11-12
Populate macOS SDKs from `joseluisq/macosx-sdks`

jaimergp:
Checklist
* [X] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

Comes from https://github.com/conda-forge/osx-sysroot-feedstock/pull/6#issuecomment-3518788930

cc @h-vetinari

### DIFF
--- a/recipe/CONTRIBUTING.md
+++ b/recipe/CONTRIBUTING.md
@@ -10,6 +10,5 @@ Usually fetched from https://github.com/joseluisq/macosx-sdks.
     1. Define `MACOSX_SDK_VERSION` in `conda_build_config.yaml`; e.g. `MACOS_SDK_VERSION: ["26.1"] # [osx]`.
     2. Add a dummy variable using in it `meta.yaml`; e.g. `{% set _dummy = MACOSX_SDK_VERSION %}`.
     3. Pick `osx` and a single Python version, and skip the rest; e.g. `skip: true # [not (osx and py==313)]`
-    4. Patch `download_osx_sdk.sh` so the line `if [[ ! -d ${CONDA_BUILD_SYSROOT} ]]; then` is always true; e.g. `if [[ 1 == 1 ]]; then`.
-    5. Rerender and push the commits.
-    6. When everything passes, revert steps 1-4 and rerender again so it's ready for review/merge.
+    4. Rerender and push the commits.
+    5. When everything passes, revert steps 1-4 and rerender again so it's ready for review/merge.


recipe/download_osx_sdk.sh
```diff
@@ -26,24 +26,35 @@ export CONDA_BUILD_SYSROOT="${OSX_SDK_DIR}/MacOSX${MACOSX_SDK_VERSION}.sdk"
 if [[ ! -d ${CONDA_BUILD_SYSROOT} ]]; then
```
h-vetinari: We should test this here; you could manually set a few different SDK versions in the variant configs
```suggestion
if [[ 1 == 1 ]]; then
```

jaimergp: Looks like it's working.

isuruf (changes requested):
We don't need the minor version from 11 onwards. The equivalent is the patch version in pre-11 versions.

h-vetinari:
> We don't need the minor version from 11 onwards. The equivalent is the patch version in pre-11 versions.

In that case we should lint when someone sets `c_stdlib_version` or `MACOSX_SDK_VERSION` to something like
```
13.0
13.1
13.2
13.3
```
and just allow 11, 12, 13, 14, 15, 26.

On the other hand, macOS will change the vendored LLVM version within patch versions, [for example](https://github.com/llvm/llvm-project/blob/fa417d78b18f5a12e2be727efd2928cf775fe3ac/libcxx/utils/libcxx/test/features/availability.py#L50-L53), MacOS 14.0 - 14.3 is based on LLVM 16, whereas 14.4+ is based on LLVM 17. Likewise with macOS 15.0 - 15.3 (LLVM 18) vs. 15.4+ (LLVM 19). I could see it happening that these differences are observable in the SDK, and potentially (in some cases) on user systems too (while we ship our own `libc++`, we rely on the system `libc++abi` coming from the vendored LLVM), such that there might be cases where `__osx >=14.4` is justified, but not `__osx >=14`.

Given that it doesn't cost a lot to provide the minor versions here (just a few more hashes to specify), I think we might want to err on the side of caution and provide them? It'd certainly consider it more intuitive (though it's also possible that this is just habit).

isuruf:
That's about libc++ which we don't care about since we provide it ourselves.

h-vetinari:
I addressed exactly that point: 

> while we ship our own `libc++`, we rely on the system `libc++abi` coming from the vendored LLVM

If there's an ABI break in `libc++abi` between v16 & v17 (or v18 and v19, etc.), it'd be observable within a macOS major-line. Something like [this patch](https://github.com/conda-forge/libcxx-feedstock/blob/main/recipe/patches/0001-Fix-ABI-compatibility-with-system.patch) we're already carrying would then need to be applied for `__osx >=N.4` instead of for `__osx >=N`

h-vetinari:
> That's just a theoretical issue that we can handle when it comes to it.

Fine by me.

> Having SDKs populated for minor versions will not miraculously handle it.

Agreed, if at all it's about `__osx` resp. `c_stdlib_version`, but there's a desired symmetry in how we handle `c_stdlib_version` and `MACOSX_SDK_VERSION`, which is at least partially what this PR is about.

recipe/download_osx_sdk.sh
```diff
@@ -24,29 +24,42 @@ fi
 export CONDA_BUILD_SYSROOT="${OSX_SDK_DIR}/MacOSX${MACOSX_SDK_VERSION}.sdk"
```
jaimergp: If we only use major versions, local copies might stay in older minor releases. Not sure if we want or need to do something about it.

jaimergp: Changed this to use the actual version instead.

h-vetinari:
> For eg: 15.0 to 15.5 adds new symbols

Recent example where this is relevant: https://github.com/conda-forge/blas-feedstock/pull/158

h-vetinari:
> It looks like macosx adds new symbols on minor versions. [...] So, having the ability to download minor versions is useful.

The question is if we'd ever need 15.0 if we can use 15.5? If we map `'15'` to `15.<latest_minor_available_in_ci_setup>`, then the major-only theme should still work.

h-vetinari:
> It looks like macosx adds new symbols on minor versions

Perhaps the LAPACK case it not representative of what I was thinking about when responding above (i.e. that 15.5 is always preferable to 15.0). When I think "MACOSX_SDK", it's usually the thing we can raise arbitrarily without impacting the deployment target, but I'm not sure if that's actually true for the `$NEWLAPACK` support?

recipe/CONTRIBUTING.md
```diff
@@ -0,0 +1,15 @@
...
+
+Usually fetched from https://github.com/joseluisq/macosx-sdks.
+
+- Add version and its checksum to `download_osx_sdk.sh`.
+- Add version to `test_osx_sdk.sh` versions array to verify the download.
+- Temporarily test it works:
+    1. Define `MACOSX_SDK_VERSION` in `conda_build_config.yaml`; e.g. `MACOS_SDK_VERSION: ["26.1"] # [osx]`.
+    2. Add a dummy variable using in it `meta.yaml`; e.g. `{% set _dummy = MACOSX_SDK_VERSION %}`.
+    3. Pick `osx` and a single Python version, and skip the rest; e.g. `skip: true # [not (osx and py==313)]`
+    4. Patch `download_osx_sdk.sh` so the line `if [[ ! -d ${CONDA_BUILD_SYSROOT} ]]; then` is always true; e.g. `if [[ 1 == 1 ]]; then`.
```
isuruf: This is unnecessary after https://github.com/conda-forge/conda-smithy/pull/2430

isuruf:
Thanks

PR #408, opened 2025-11-16
 Delete clang(s) in windows runner images that get prioritized over our own compilers

h-vetinari:
The windows runner image started shipping clang binaries on a path that gets prioritized over compilers from `build:`. This leads to various work-arounds (not yet super-widespread because clang is not the default compiler on windows), for example
* [clangdev](https://github.com/conda-forge/clangdev-feedstock/blob/601374190c2b542ad59f6a5c4bb393d3869e64d0/recipe/meta.yaml#L499-L500)
* [openmp](https://github.com/conda-forge/openmp-feedstock/blob/5c874b6f5ee382950f0f64515f289096a48d8511/recipe/bld.bat#L13-L14)
* [cython-blis](https://github.com/conda-forge/cython-blis-feedstock/blob/dbdbecefb9684617918a3d885ba12fab8d1690d3/recipe/bld.bat#L3-L5)

There's probably several other feedstocks where this simply did not become obvious yet - absent some other constraints (e.g. cython-blis needs clang <19 and fails otherwise), this needs fairly close study of the logs to find that the wrong clang is used, which is of course a risk in the sense that we're suddenly building stuff with a toolchain whose configuration we don't control (for homogeneity of various flags, etc.).

Rather than leaving this footgun lying around, we should remove these vendored clangs in all our CI runs. If people want to use clang on windows, they should add it to the respective `build:` environment, as with everything else.

In addition to the clang vendored by MSVC, the runner images _also_ add LLVM [directly](https://github.com/actions/runner-images/blob/main/images/windows/scripts/build/Install-LLVM.ps1) from choco. We remove both variants here, including the respective path for the just-released VS2026, which will show up in the images quite soon.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "4.23.1" %}
+{% set version = "4.23.2" %}
 {% set build = 0 %}
 
 {% set cuda_compiler_version = cuda_compiler_version or "None" %}
--- a/recipe/run_conda_forge_build_setup_win.bat
+++ b/recipe/run_conda_forge_build_setup_win.bat
@@ -78,6 +78,18 @@ if defined CI (
     DEL C:\Windows\System32\msmpires.dll || (Echo Ignoring failure to delete C:\Windows\System32\msmpires.dll)
 )
 
+:: MSVC vendors clang, which gets picked up before our own compilers, so we delete it
+:: Additionally, the chocolatey-based installation of LLVM in the images may bring in yet another LLVM, c.f.
+:: https://github.com/actions/runner-images/blob/main/images/windows/scripts/build/Install-LLVM.ps1
+:: and the end of `chocolateyinstall.ps1` in https://community.chocolatey.org/packages/llvm/20.1.8#files
+if defined CI (
+    RMDIR /s /q "C:\Program Files\LLVM" || (Echo Ignoring failure to delete C:\Program Files\LLVM)
+    RMDIR /s /q "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\Llvm" ^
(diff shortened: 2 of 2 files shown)


h-vetinari:
> When I built some workarounds, I first also thought of deleteing the binaries but as this comes with MSVC, we would probably face the same issue on local builds, too? There we shouldn't delete stuff from user's MSVC installation and thus I think we will face the same issues there as we would face in CI.

That's why the whole thing is guarded by an `if defined CI`. Local builds IMO aren't anywhere near as important - if someone runs into such a situation on a local build, they can make the choice if they live with the vendored clang or whether they want to change their local MSVC install. That's a negligible price to pay for ensuring that CI runs properly.

Obviously it'd be nice if things just worked the same way everywhere, but the way the runner images are populated and windows does its default path resolution just make this impossible IMO.

h-vetinari:
> The 3 workarounds mentioned can be removed by making vc-feedstock fix the PATH after activating the compilers.

Sure, we can fix this in compiler activation, but one of the points you've put substantial attention on was to keep our compilers working without activation too. This happens while building the LLVM stack of course, but also just for recipes that end up doing
```
build:
  - clang  # [win]
```
without any activated compilers. This works for the most part, but is a footgun IMO. We don't put GCC in the docker-images either (and then rely on avoiding it through compiler activation), for me this is similar, except we can't package our own images for windows.

isuruf:
> Sure, we can fix this in compiler activation, but one of the points you've put substantial attention on was to keep our compilers working without activation too.

The current issue has nothing to do with using compilers without activation. The issue is that activating `vs2022_win-64` puts the path of MSVC's clang to the front. If `vs2022_win-64` was not activated, conda would be in the front and therefore no issue there.

h-vetinari:
> We call vcvarsall.bat which adds `%VSINSTALLDIR%` related paths to the front. i.e. they get more priority than `%BUILD_PREFIX%`, `%PREFIX%` paths. The solution is to move them to the front after vcvarsall.bat call.

This was done in https://github.com/conda-forge/vc-feedstock/pull/115; hope that fixes things; in the meantime I'm closing this

PR #417, opened 2026-04-21
Allow C:\bld and D:\bld for pagefile setting

mgorny:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

Per the discussion in conda-forge/conda-smithy#2529.

PR #418, opened 2026-05-14
Fix Windows pagefile drive expansion

RSchwan:
Checklist

Fix Windows pagefile setup when `CONDA_BLD_PATH` is `C:\bld` or `D:\bld`.

The previous code assigned `PageFileDrive` inside a parenthesized batch block and then read it later in the same block with `%PageFileDrive%`. In `cmd.exe`, percent variables are expanded before the whole block executes, so the value was still empty when passed to `SetPageFileSize.ps1`. This resulted in logs like `Setting pagefile size to 16GB in` and a PowerShell `The handle is invalid` error.

This moves the `PageFileDrive` calculation outside the block, fixes the guard to use `%PageFileDrive%`, and handles both normal and YAML-escaped path forms with a trailing backslash.

Example failure:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1520770&view=logs&j=78751870-5363-50c8-9028-5b2786f088ed&t=4b9f6b4c-d277-520a-727a-905d6908cd61&l=347

```text
CONDA_BLD_PATH=D:\\bld\\; Setting pagefile size to 16GB in
[...]
Exception calling "SetPageFileSize" with "3" argument(s): "The handle is invalid"
At D:\Miniforge\Scripts\SetPageFileSize.ps1:197 char:1
+ [Util.PageFile]::SetPageFileSize($minimumSize, $maximumSize, $diskRoo ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : Win32Exception
```

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "conda-forge-ci-setup" %}
-{% set version = "4.24.3" %}
-{% set build = 1 %}
+{% set version = "4.24.4" %}
+{% set build = 0 %}
 
 {% set cuda_compiler_version = cuda_compiler_version or "None" %}
 {% if cuda_compiler_version == "None" %}
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "conda-forge-ci-setup" %}
 {% set version = "4.24.3" %}
-{% set build = 0 %}
+{% set build = 1 %}
 
 {% set cuda_compiler_version = cuda_compiler_version or "None" %}
 {% if cuda_compiler_version == "None" %}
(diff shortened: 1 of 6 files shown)


RSchwan:

recipe/meta.yaml
```diff
@@ -1,6 +1,6 @@
 {% set name = "conda-forge-ci-setup" %}
 {% set version = "4.24.3" %}
-{% set build = 0 %}
+{% set build = 1 %}
```
h-vetinari: for the CI setup we generally don't do build numbers; just increment the patch version.

h-vetinari (commented):
Let's take this as an improvement over #417, where there was some breakage that has not been followed up upon AFAICT. We might have to revert both PRs, but I still don't really know or understand what issues @mgorny had run into

mgorny:
Oh, thanks for figuring this out. CMD is truly a mystery to me.

h-vetinari:
Thanks for the PR @RSchwan! 🙏

PR #423, opened 2026-07-17
Do not override BUILD_PLATFORM if already defined

jaimergp:
Goes with https://github.com/conda-forge/conda-smithy/pull/2622, where we may define BUILD_PLATFORM before it reaches ci-setup. In that case, do not override.

cc @isuruf


beckermr (commented):
Do we need to check the `.sh` script too?

beckermr (approved):

h-vetinari:
I think you forgot to bump the version here

PR #420, opened 2026-07-03
set CONDA_OVERRIDE_CUDA_ARCH when cuda_arch_version is set in conda_forge_config.yaml

carterbox:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

In this MR, conda-forge-ci-setup is extended to set `CONDA_OVERRIDE_CUDA_ARCH` in the CI similarly to how `CONDA_OVERRIDE_CUDA` is set.

This is needed in order to build packages which require `__cuda_arch`. Otherwise, during the test phase, the package being built will be uninstallable.

PR #427, opened 2026-08-13
Add nvidia-virtual-packages as a dependency / Fix CONDA_OVERRIDE_CUDA_ARCH for Windows

carterbox:
Checklist

This PR is a followup to #420 which fixes a few bugs in the Windows activation script for CONDA_OVERRIDE_CUDA_ARCH, adds a test for CONDA_OVERRIDE_CUDA_ARCH, and fixes a broken skip for cuda build variants of this package.

The reason these bugs were not caught before is that we didn't have active tests, and I only had one test recipe using __cuda_arch which was using Linux + rattler-build. These bug fixes should allow __cuda_arch to be available for recipes building on Windows and recipes building with conda-build (which requires a plugin for __cuda_arch support).

A summary of the bugs fixed is as follows:

1. nvidia-virtual-packages is required in the CI setup environment in order for conda/conda-build to provide __cuda_arch
3. Writes to `conda-forge-ci-setup-activate.bat` were being reset after CONDA_OVERRIDE_CUDA(_ARCH) were added to the file.
4. CUDA_ARCH_VERSION was defined in a local context only
5. conda-smithy was pruning cuda_compiler_version_min from the variant files causing the skip statement to be true at build time.


recipe/run_conda_forge_build_setup_win.bat
carterbox, line 150: Activate script is overwritten (>) here, so we need to write CONDA_OVERRIDE_CUDA(_ARCH) after this point instead of before.

recipe/meta.yaml
carterbox, line 21: This skip was evaluating to true because `cuda_compiler_version_min` was undefined.

recipe/meta.yaml
```diff
@@ -95,16 +96,22 @@ requirements:
     - conda-forge-metadata >=0.9.2
     - conda-package-handling >=2.3.0
     - rattler-build-conda-compat >=0.0.2,<2.0.0a0
+    - nvidia-virtual-packages
```
carterbox, line 99: I believe this is the correct place to add a dependency on nvidia-virtual-packges. The other place I considered was in the conda-smithy template.

jaimergp: Shouldn't this go in `conda-feedstock` instead? 🤔

carterbox: No? We don't require that users have `nvidia-virtual-packages` to use our channel.

`__cuda_arch` is an opt-in feature. If we made it an optional dependency for only CUDA variants of this package, would that work correctly? By that I mean do CUDA build runners only consumer CUDA variants of conda-forge-ci-setup or are the variants just a way to run the tests against different environments?

carterbox: I think `nvidia-virtual-packages` is a bit different than `archspec` because every user will have a CPU, but only some users will have a CUDA device.

isuruf: Sure, but it's a very small package to have as a dependency of conda.

jaimergp: > but only some users will have a CUDA device.

`conda` already carries the implementation for `__cuda` and it's ok. Actually, I always thought that the `__cuda` implementation would end up moving to `nvidia-virtual-packages`, hence becoming a default dependency.

While I understand your point, I don't think it's necessary to optimize in that way.

jaimergp: > By that I mean do CUDA build runners only consumer CUDA variants of conda-forge-ci-setup or are the variants just a way to run the tests against different environments?

I think this is just for testing, I don't see variants for CUDA in https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fconda-forge-ci-setup-4.26.2-py314h36c64d2_100.conda (only Python)

jaimergp: That said, we can start with this PR like this, and when irremediably this dependency gets added to conda (I assume in 26.9, once we have dealt with https://github.com/conda/conda/pull/16509), we can remove it from here. WDYT?

carterbox: - [ ] https://github.com/conda-forge/conda-feedstock/pull/315

carterbox:

recipe/meta.yaml
carterbox: Rebasing this PR since I'm not getting any attention from the conda maintainers.

carterbox: Adding the package here, then adding to conda later is the more conservative approach. We can do both, and the dependency here is unconstrained, so there will not be conflicts.

carterbox:
I'm going to merge this tomorrow unless there is further feedback.

carterbox:

PR #432, opened 2026-08-30
read c_stdlib_version instead of MACOSX_DEPLOYMENT_TARGET

h-vetinari:
Back during https://github.com/conda-forge/conda-forge.github.io/issues/2102, we kept `MACOSX_DEPLOYMENT_TARGET` in place, so as to not also have to do surgery on ci-setup, especially while the transition was ongoing. This required ugly hacks in smithy to keep working, e.g. https://github.com/conda-forge/conda-smithy/pull/1889

Now that 2+ years have gone by, it's fair to expect that all feedstocks (and their osx variant configs) have a c_stdlib_version, so we can change the CI setup to reflect that, which will both enable cleaning up smithy, which will also solve issues like what https://github.com/conda-forge/conda-smithy/pull/2666 is trying to solve.

The other occurrences of `MACOSX_DEPLOYMENT_TARGET` in that file haven't been touched, because we still want to set the corresponding environment variable, in case some build systems depend on that.

### DIFF
--- a/recipe/download_osx_sdk.sh
+++ b/recipe/download_osx_sdk.sh
@@ -2,7 +2,7 @@ set -e
 
 # We don't change the default here to a newer SDK to ensure that old, non-rerendered feedstock keep working.
 if [ -f ${CI_SUPPORT}/${CONFIG}.yaml ]; then
-   export MACOSX_DEPLOYMENT_TARGET=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value MACOSX_DEPLOYMENT_TARGET.0 10.9)
+   export MACOSX_DEPLOYMENT_TARGET=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value c_stdlib_version.0 10.9)
 fi
 
 export MACOSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET:-10.9}


hmaarrfk:
Thanks for looking into this. This makes sense.

jaimergp:
What's the migration path here? Would a simple rerender suffice? I see [a lot of hits](https://github.com/search?type=code&q=org%3Aconda-forge+path%3A.ci_support%2Fosx_*.yaml+MACOSX_DEPLOYMENT_TARGET+NOT+c_stdlib_version%3A) with `MACOSX_DEPLOYMENT_TARGET` but no `c_stdlib_version`.

h-vetinari:
A lot of these simply haven't been rerendered in years, or not responded to the linter directive to add `{{ stdlib("c") }}`, or simply don't have a compiler (where the macOS SDK becomes necessary). After 2+ years, I'm fine with breaking them if they can't respond to a lint in that time.

At least, I reluctantly prefer that over adding yet more hacks to solve issues around needing to provide dual builds against newer macOS versions for some feedstocks (i.e. what https://github.com/conda-forge/conda-smithy/pull/2666 was trying to achieve).

recipe/download_osx_sdk.sh
```diff
@@ -2,7 +2,7 @@ set -e
 
 # We don't change the default here to a newer SDK to ensure that old, non-rerendered feedstock keep working.
 if [ -f ${CI_SUPPORT}/${CONFIG}.yaml ]; then
-   export MACOSX_DEPLOYMENT_TARGET=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value MACOSX_DEPLOYMENT_TARGET.0 10.9)
+   export MACOSX_DEPLOYMENT_TARGET=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value c_stdlib_version.0 10.9)
 fi
```
carterbox: You could add an error which forces migration to the new approach? e.g.

```bash
if [[ "${mdt}" != "0" && "${csv}" == "0" ]]; then
    echo "ERROR: config has MACOSX_DEPLOYMENT_TARGET but no c_stdlib_version."
    echo "       Add {{ stdlib('c') }} to the recipe and rerender."
    exit 1
fi
```

That's probably better than silently degrading to deployment target 10.9?

carterbox (changes requested):
`MACOSX_DEPLOYMENT_TARGET` is also extracted from the config file in the linux setup in order to decide whether to call the download script. That logic needs to be updated too.

https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/f7bd98b4cfe77fbfbbb4af1741ba1f7ab9eec60b/recipe/run_conda_forge_build_setup_linux#L97-L104

h-vetinari:
Thanks for the review; makes sense, adopted!

recipe/run_conda_forge_build_setup_linux
```diff
@@ -95,8 +95,14 @@ SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
 
 if [ -f ${CI_SUPPORT}/${CONFIG}.yaml ]; then
+    csv=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value c_stdlib_version.0 0)
     mdt=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value MACOSX_DEPLOYMENT_TARGET.0 0)
     msv=$(cat ${CI_SUPPORT}/${CONFIG}.yaml | shyaml get-value MACOSX_SDK_VERSION.0 0)
+    if [[ "${mdt}" != "0" && "${csv}" == "0" ]]; then
+        echo "ERROR: config has MACOSX_DEPLOYMENT_TARGET but no c_stdlib_version."
+        echo "       Add {{ stdlib('c') }} to the recipe and rerender."
+        exit 1
+    fi
```
carterbox: I think this check is redundant with the identical check in `download_osx_sdk.sh` because `download_osx_sdk.sh` is run whenever `"${mdt}" != "0"`, so whenever `mdt` is set, then the check will run.

carterbox (approved):
LGTM after the redundant check is removed or the author provides any reason why they want to check in both places.

